PR #168, opened 2021-05-24
[gitlab] add identifier field

### LINKED ISSUE #166: [gitlab] Release fails if Project Fork exist
### Background 
My company uses an internal GitLab instance, and we are working on a new project, which has the same name as an already existing one that we are replacing. I also have a fork of the original one, which I use to propose MRs to the maintaining team. Hence, my account sees three repositories with the same name.

### Error
Under these circumstances, if I use the CLI approach to create a release, I get a RestAPIException [from Feign](https://github.com/jreleaser/jreleaser/blob/main/sdks/java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/ClientUtils.java#L81). 

My understanding of what has happened: Since GitLabs API requires using the unique project IDs for a call, JReleaser obtains the Project by searching for a project with the `release.gitlab.name` value. Since my account sees three of these, it fetches the wrong one. 

https://github.com/jreleaser/jreleaser/blob/65b0e520d3bcc6c129e725ae42d65a1bfa6b55aa/sdks/gitlab-java-sdk/src/main/java/org/jreleaser/sdk/gitlab/Gitlab.java#L105-L117

It now uses the wrong project id, and since the desired branch does not exist (master vs. main) GitLab answers with a 400. 

### Environment Information

- **Operating System**:  Mac
- **JReleaser Version:** v0.3
- **JDK Version:** The Homebrew provided one, I believe. 

### Ideas:
- Add a config parameter for GitLab releases to define the project ID explicitly.
- Possibly, this would allow to not rely on personal Tokens but use project-specific ones. Teams could store this Token value in a standard secure enclave  (direnv to a secure, shared team folder, for instance) such that each team member can create a release without providing an additional token. The technical user behind a project token is not allowed to query the getProject API.
- Build something that catches the RestAPIException and allows to handle this error gracefully.

DennisRippinger:
This PR adds the possibility to define the GitLab-specific project IDs. This closes #166.

The mentioned idea that GitLab returns multiple results and JReleaser picks the wrong one is not valid. I discovered that the query returns one result item and prefers a project in the personal space that has the same name, aka a fork. Only if the name is not an exact match, the query returns multiple results. Anyhow, the proposed solution should provide a remedy to that issue. 

I implemented as proposed in the issue. But I think this PR could do better. The SDK GitLab class uses the Model class and spreads the parameter to the submethods. With `identifier this grows even further. If the methods would directly accept git model class, `getProject` could compare the desired path of owner and name if an identifier is not present. That would allow providing a meaningful warning.

### DIFF
--- a/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/GitlabImpl.groovy
+++ b/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/GitlabImpl.groovy
@@ -36,6 +36,7 @@ class GitlabImpl extends AbstractGitService implements Gitlab {
     final ChangelogImpl changelog
     final MilestoneImpl milestone
     final CommitAuthorImpl commitAuthor
+    final Property<String> identifier
 
     @Inject
     GitlabImpl(ObjectFactory objects) {
@@ -44,6 +45,8 @@ class GitlabImpl extends AbstractGitService implements Gitlab {
         changelog = objects.newInstance(ChangelogImpl, objects)
         milestone = objects.newInstance(MilestoneImpl, objects)
         commitAuthor = objects.newInstance(CommitAuthorImpl, objects)
+
+        identifier = objects.property(String).convention(Providers.notDefined())
     }
 
     @Override
@@ -61,6 +64,7 @@ class GitlabImpl extends AbstractGitService implements Gitlab {
         if (changelog.isSet()) service.changelog = changelog.toModel()
         if (milestone.isSet()) service.milestone = milestone.toModel()
         if (commitAuthor.isSet()) service.commitAuthor = commitAuthor.toModel()
+        if (identifier.present) service.identifier = identifier.get()
         service
     }
 }
--- a/plugins/jreleaser-maven-plugin/src/main/java/org/jreleaser/maven/plugin/internal/JReleaserModelConverter.java
+++ b/plugins/jreleaser-maven-plugin/src/main/java/org/jreleaser/maven/plugin/internal/JReleaserModelConverter.java
@@ -166,6 +166,7 @@ public final class JReleaserModelConverter {
     private static org.jreleaser.model.Gitlab convertGitlab(Gitlab gitlab) {
         if (null == gitlab) return null;
         org.jreleaser.model.Gitlab g = new org.jreleaser.model.Gitlab();
+        g.setIdentifier(gitlab.getIdentifier());
         convertGitService(gitlab, g);
         return g;
     }
--- a/sdks/gitlab-java-sdk/src/main/java/org/jreleaser/sdk/gitlab/Gitlab.java
+++ b/sdks/gitlab-java-sdk/src/main/java/org/jreleaser/sdk/gitlab/Gitlab.java
@@ -38,6 +38,7 @@ import org.jreleaser.sdk.gitlab.api.Release;
 import org.jreleaser.sdk.gitlab.api.User;
 import org.jreleaser.util.CollectionUtils;
 import org.jreleaser.util.JReleaserLogger;
+import org.jreleaser.util.StringUtils;
 
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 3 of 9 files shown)


plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/GitlabImpl.groovy
```diff
@@ -44,6 +45,8 @@ class GitlabImpl extends AbstractGitService implements Gitlab {
         changelog = objects.newInstance(ChangelogImpl, objects)
         milestone = objects.newInstance(MilestoneImpl, objects)
         commitAuthor = objects.newInstance(CommitAuthorImpl, objects)
+
+        identifier = objects.property(String).convention(Providers.notDefined())
```
aalmiray, line 49: The `identifier` field is missing in `toModel()`

plugins/jreleaser-maven-plugin/src/main/java/org/jreleaser/maven/plugin/internal/JReleaserModelConverter.java
```diff
@@ -220,6 +220,14 @@ private static void convertGitService(GitService service, org.jreleaser.model.Gi
         s.setMilestone(convertMilestone(service.getMilestone()));
         s.setConnectTimeout(service.getConnectTimeout());
         s.setReadTimeout(service.getReadTimeout());
+
+        convertGitlabSpecific(service, s);
+    }
+
+    private static void convertGitlabSpecific(GitService service, org.jreleaser.model.GitService s) {
```
aalmiray, line 227: This method may be removed and its contents added to `convertGitlab()`

DennisRippinger: Ah, there 😅

plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/GitlabImpl.groovy
DennisRippinger: My first change to non-gradle groovy code. I hope it fits now.

plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/GitlabImpl.groovy
```diff
@@ -61,6 +64,7 @@ class GitlabImpl extends AbstractGitService implements Gitlab {
         if (changelog.isSet()) service.changelog = changelog.toModel()
         if (milestone.isSet()) service.milestone = milestone.toModel()
         if (commitAuthor.isSet()) service.commitAuthor = commitAuthor.toModel()
+        service.identifier = identifier
```
aalmiray, line 67: Good. However this call might produce an error if the value has not been initialized. There are a couple of ways to prevent that, the rest of the code does it like this `if (identifier.present) service.identifier = identifier.get()`.

sdks/gitlab-java-sdk/src/main/java/org/jreleaser/sdk/gitlab/Gitlab.java
```diff
@@ -102,24 +103,14 @@
...
-
-        return projects.get(0);
+    Project findProject(String projectName, String identifier) throws RestAPIException {
+        return getProject(projectName, identifier);
     }
 
-    Optional<Milestone> findMilestoneByName(String owner, String repo, String milestoneName) throws IOException {
-        logger.debug("Llookup milestone '{}' on {}/{}", milestoneName, owner, repo);
+    Optional<Milestone> findMilestoneByName(String owner, String repo, String identifier, String milestoneName) throws IOException {
+        logger.debug("Lookup milestone '{}' on {}/{}", milestoneName, owner, repo);
```
aalmiray, line 111: Please change this to lower case `lookup`.

aalmiray:
Thanks a lot! 😁

PR #229, opened 2021-06-23
[maven] Using element name to set Distribution name. Fixes #228

### LINKED ISSUE #228: [maven] Documentation for Distributions is missing the name attribute
Documentation for distributions via Maven does not mention the `name` element.
This does not work:
```
<jreleaser>
  <distributions>
    <app>
      <artifacts>...</artifacts>
    </app>
  </distributions>
</jreleaser>
```

but this does:
```
<jreleaser>
  <distributions>
    <app>
      <name>myapp</name>
      <artifacts>...</artifacts>
    </app>
  </distributions>
</jreleaser>
```

Happy to submit a PR if needed.

jruaux:
Fixes #228

When configuring the Maven plugin, Distribution name is not taken from the XML element name.

### DIFF
--- a/plugins/jreleaser-maven-plugin/src/main/java/org/jreleaser/maven/plugin/Jreleaser.java
+++ b/plugins/jreleaser-maven-plugin/src/main/java/org/jreleaser/maven/plugin/Jreleaser.java
@@ -17,9 +17,8 @@
  */
 package org.jreleaser.maven.plugin;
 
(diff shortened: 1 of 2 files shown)


aalmiray:
I totally missed that distributions were set as a `List` instead of a `Map` as the other entities that follow the same named pattern. Good catch!

aalmiray:
Thank you for the PR!

PR #535, opened 2021-11-17
Translation in Russian provided.

dalexandrov:
:)


aalmiray:
Thank you so much @dalexandrov! Though something is strange with the CLI bundles as all cyrillic characters appear as ???. The core bundle looks OK thoug
ru
h.

aalmiray:
I think the problem is that both Banner and Messages are encoded with ISO 8859-1 instead of UTF-8. Might be a side effect of creating the file with IntelliJ. You can create a text file with vi for example, copy the content into it, then mv the new file into the previous one. That should take care of the problem 😄

dalexandrov:
Now it should be fine :)

aalmiray:
Thanks! 😄

PR #594, opened 2021-12-10
Insert JReleaser version and generation timestamp to generated files

### LINKED ISSUE #593: [templates] Add a comment on generated templates stating JReleaser version used
Besides the current JReleaser version it could also add a timestamp

Idea by @mthmulders https://github.com/jreleaser/jreleaser/discussions/588#discussioncomment-1784891

mthmulders:
Resolves #593.

### Context
See #593 and #588. The idea is that a comment in generated files should point users to the fact that the generated file might be outdated, comparing to the version of JReleaser they currently use.

There are a few distribution templates that are JSON files - as far as I know, there is no way to insert a comment in JSON files.

I've also skipped the announcement templates, because
a) they are not meant for computer processing, but rather for reading by humans.
b) they do not have a predefined format that specifies what a comment should look like.

### Checklist


core/jreleaser-templates/src/main/java/org/jreleaser/templates/VersionDecoratingWriter.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jreleaser.model.JReleaserVersion;
+import org.jreleaser.util.JReleaserLogger;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.Writer;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class VersionDecoratingWriter extends BufferedWriter {
```
aalmiray, line 29: @mthmulders please add an `@author` tag with your name 😏

mthmulders: I didn't want to boast too much... :-)

aalmiray:
Thank you! 😄

PR #695, opened 2022-01-24
[docs] Update Jbang instructions

quintesse:
Removed "Required Java 8" from the Jbang instructions as this not really necessary given the fact Jbang can install Java when needed.


aalmiray:
Thanks! 😄

PR #767, opened 2022-02-20
[snap] versions need to be strings

### LINKED ISSUE #766: [Snapcraft] versions need to be strings
### Discussed in https://github.com/jreleaser/jreleaser/discussions/765

<div type='discussions-op-text'>

<sup>Originally posted by **mthmulders** February 20, 2022</sup>
During my last release with JReleaser (v0.9.1), the Snapcraft package was initially rejected with

> Issues while validating snapcraft.yaml: The 'version' property does not match the required schema: snap versions need to be strings. They must also be wrapped in quotes when the value will be interpreted by the YAML parser as a non-string. Examples: '1', '1.2', '1.2.3', git (will be replaced by a git describe based version string).

I've [monkey-patched the YAML file](https://github.com/mthmulders/mcs-snap/commit/2fd6ea8ab1130b53e6de8a9faf38cde548ea7116) so the version number is enclosed in quotes, and then the Snapcraft package was built fine.

I've done a diff between JReleaser 0.9.1 and HEAD and it seems none of the Snapcraft related code has changed since - hence I suspect the issue will still be there. If so, would fixing this be considered for JReleaser 1.0.0? Maybe I can contribute?</div>

mthmulders:
Fixes #766 

### Context

A recent Snapcraft build of mcs v0.2 failed with 

> Issues while validating snapcraft.yaml: The 'version' property does not match the required schema: snap versions need to be strings. They must also be wrapped in quotes when the value will be interpreted by the YAML parser as a non-string. Examples: '1', '1.2', '1.2.3', git (will be replaced by a git describe based version string).

### Checklist
- [X] Update [documentation](https://github.com/jreleaser/jreleaser.github.io) when applicable.

### DIFF
--- a/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/binary/snap/snap/snapcraft.yaml.tpl
+++ b/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/binary/snap/snap/snapcraft.yaml.tpl
@@ -1,6 +1,6 @@
 # {{jreleaserCreationStamp}}
 name: {{snapPackageName}}
-version: {{projectVersion}}
+version: "{{projectVersion}}"
 summary: {{projectDescription}}
 description: {{projectLongDescription}}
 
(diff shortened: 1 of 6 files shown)


aalmiray:
Thanks! 😄

PR #797, opened 2022-03-20
Not parseable and unmatched tags are now excluded from tags

### LINKED ISSUE #796: [changelog] Wrong commits selected when not parseable tags found
### Task List

#### Case 1 - tags do not match the tag name
1. Have a repo with tags "v1.0.0", "v2.0.0"
2. Use as tagName release-{{projectVersion}}
3. JRELEASER_PROJECT_VERSION=2.2.0 jrelease changelog

#### Case 2 - tags that are matched due to (.*) but not parseable
1. Have a repo with tags "v1.0.0", "v2.0.0"
2. Use as tagName {{projectVersion}}
3. JRELEASER_PROJECT_VERSION=2.2.0 jrelease changelog

### Expected Behaviour
I expect that all commits will be used until the head.

### Actual Behaviour

tag v1.0.0 will be used as a start point

### Environment Information

- **Operating System**: Windows
- **JReleaser Version:** v1.0.0-M3
- **JDK Version:** 1.8

mfvitale:
Fixes #796

### Context

Resolve this issue #796

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,10 +90,10 @@ subprojects {
         testImplementation "org.junit.jupiter:junit-jupiter:$junit5Version"
         testImplementation "org.junit.jupiter:junit-jupiter-params:$junit5Version"
         testImplementation "org.hamcrest:hamcrest-library:$hamcrestVersion"
-        testImplementation "org.mockito:mockito-core:3.+"
-        testImplementation 'org.mockito:mockito-junit-jupiter:3.+'
+        testImplementation "org.mockito:mockito-core:$mockitoVersion"
+        testImplementation "org.mockito:mockito-junit-jupiter:$mockitoVersion"
 
-        testCompile "org.assertj:assertj-core:3.11.1"
+        testImplementation "org.assertj:assertj-core:$assertjVersion"
 
         testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junit5Version"
     }
--- a/build.gradle
+++ b/build.gradle
@@ -90,6 +90,10 @@ subprojects {
         testImplementation "org.junit.jupiter:junit-jupiter:$junit5Version"
         testImplementation "org.junit.jupiter:junit-jupiter-params:$junit5Version"
         testImplementation "org.hamcrest:hamcrest-library:$hamcrestVersion"
+        testImplementation "org.mockito:mockito-core:3.+"
+        testImplementation 'org.mockito:mockito-junit-jupiter:3.+'
+
+        testCompile "org.assertj:assertj-core:3.11.1"
 
         testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junit5Version"
     }
--- a/sdks/git-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
+++ b/sdks/git-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
@@ -50,6 +50,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -77,9 +78,9 @@ public class ChangelogGenerator {
     private static final String UNCATEGORIZED = "<<UNCATEGORIZED>>";
     private static final String REGEX_PREFIX = "regex:";
 
-    private final Set<String> unparseableTags = new LinkedHashSet<>();
+    private final Set<Ref> unparseableTags = new LinkedHashSet<>();
 
-    private ChangelogGenerator() {
+    protected ChangelogGenerator() {
 
     }
 
@@ -138,7 +139,7 @@ public class ChangelogGenerator {
         return String.join(commitSeparator, lines);
     }
 
-    private void unparseableTag(JReleaserContext context, String tag, Exception exception) {
+    private void unparseableTag(JReleaserContext context, Ref tag, Exception exception) {
         if (!unparseableTags.contains(tag)) {
             unparseableTags.add(tag);
             context.getLogger().warn(exception.getMessage());
@@ -152,7 +153,7 @@ public class ChangelogGenerator {
             try {
                 return SemVer.of(tag);
             } catch (IllegalArgumentException e) {
-                unparseableTag(context, tag, e);
+                unparseableTag(context, ref, e);
             }
         }
         return SemVer.of("0.0.0");
@@ -165,7 +166,7 @@ public class ChangelogGenerator {
             try {
                 return JavaRuntimeVersion.of(tag);
             } catch (IllegalArgumentException e) {
-                unparseableTag(context, tag, e);
+                unparseableTag(context, ref, e);
             }
         }
         return JavaRuntimeVersion.of("0.0.0");
@@ -178,7 +179,7 @@ public class ChangelogGenerator {
             try {
                 return JavaModuleVersion.of(tag);
             } catch (IllegalArgumentException e) {
-                unparseableTag(context, tag, e);
+                unparseableTag(context, ref, e);
             }
         }
         return JavaModuleVersion.of("0.0.0");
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -90,6 +90,10 @@ subprojects {
         testImplementation "org.junit.jupiter:junit-jupiter:$junit5Version"
         testImplementation "org.junit.jupiter:junit-jupiter-params:$junit5Version"
         testImplementation "org.hamcrest:hamcrest-library:$hamcrestVersion"
+        testImplementation "org.mockito:mockito-core:3.+"
```
aalmiray, line 93: Please use explicit versions, version ranges are a bad idea.
Write down the version in `gradle.properties`.

build.gradle
```diff
@@ -90,6 +90,10 @@ subprojects {
...
+        testImplementation 'org.mockito:mockito-junit-jupiter:3.+'
+
+        testCompile "org.assertj:assertj-core:3.11.1"
```
aalmiray, line 96: Configuration name should be `testImplementation`

sdks/git-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class ChangelogGeneratorUnitTest {
```
aalmiray, line 50: I'm not a fan of using package private access unless strictly necessary. 
Please make the class, fields, and test method public.

sdks/git-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
```diff
@@ -0,0 +1,116 @@
+package org.jreleaser.sdk.git;
+
+
+import javafx.util.Pair;
```
aalmiray: This causes build failures as JavaFX is not in the classpath.
You'll have to supply your own `Pair` class or use one provided by Mockito or Junit (if any)

sdks/git-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.ListTagCommand;
+import org.eclipse.jgit.api.LogCommand;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.lib.ObjectIdRef;
+import org.eclipse.jgit.lib.Ref;
+import org.jreleaser.model.*;
```
aalmiray, line 29: Please do not use star imports

sdks/git-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.mockito.Mockito.*;
```
aalmiray, line 46: Please do not use star imports

aalmiray:
@mfvitale thanks for the PR! Left a few comments that should be addressed before merging.

mfvitale:
> @mfvitale thanks for the PR! Left a few comments that should be addressed before merging.

I don't see it.

aalmiray:
@mfvitale all comments should be visible at https://github.com/jreleaser/jreleaser/pull/797/files

mfvitale:
> @mfvitale all comments should be visible at https://github.com/jreleaser/jreleaser/pull/797/files

Nope. No comments. Only diff of changes.

aalmiray (changes requested):

aalmiray:
@mfvitale got it. Looks like I missed pushing a button.

mfvitale:
> @mfvitale got it. Looks like I missed pushing a button.

:-) Comments addressed.

aalmiray (approved):

sdks/git-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
```diff
@@ -251,12 +252,16 @@ private Version version(JReleaserContext context, Ref tag, Pattern versionPatter
         Pattern versionPattern = vp;
 
         unparseableTags.clear();
+
+        tags = tags.stream().filter(tagMatches(versionPattern)).collect(Collectors.toList());
```
aalmiray, line 256: Something tells me this will break `early-access` releases as the tag that identifies such a release will be filtered before it can be checked.

How to test this:
- add tags v1.0.0 and v2.0.0
- add a commit after tag v2.0.0
- project version should be snapshot, such as 2.1.0-SNAPSHOT
- `gitService.getEffectiveTagName(context.getModel());` should return `early-access` in this case.
- generate changelog. Should have a single commit (from step 2)

aalmiray (changes requested):

aalmiray:
@mfvitale added one more review comment

sdks/git-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
mfvitale: I'll do a test on that case.

mfvitale: I think also this 'tags.removeAll(unparseableTags);' can break. What is the expected behavior here? If the project is a snapshot you are using 'early-access' as 'effectiveTagName' but why match all tags even if the tagName pattern is not matching?

mfvitale: I want to add that this seems not working also before. The code inside the early-access branch

`tag = tags.stream()
                            .filter(ref -> !extractTagName(ref).equals(effectiveTagName))
                            .filter(ref -> currentVersion.equalsSpec(version(context, ref, versionPattern)))
                            .findFirst();`
Since in this case, the effectiveTagName is 'early-access' all tags will be filtered out.

aalmiray: Correct. I think I have a fix for the last concern you raised, by inserting one additional filter

```
                        tag = tags.stream()
                            .filter(ref -> !extractTagName(ref).equals(effectiveTagName))
                            .filter(ref -> versionPattern.matcher(extractTagName(ref)).matches())
                            .filter(ref -> currentVersion.equalsSpec(version(context, ref, versionPattern)))
                            .findFirst();
```

aalmiray: > I think also this 'tags.removeAll(unparseableTags);' can break. What is the expected behavior here? If the project is a snapshot you are using 'early-access' as 'effectiveTagName' but why match all tags even if the tagName pattern is not matching? So if I have a project in a snapshot mode I have to match again all possible tags not only the ones that matched the tagName pattern?

Correct. When a project is deemed as snapshot the `early-access` tag will compute the starting red which could be any of:

- an existing `early-access` tag
- the previously released tag, i.e. `v1.0.0`
- the start of the git history if no tags

aalmiray:
@mfvitale committed a fix based on your PR (with attribution) plus a few additional changes

https://github.com/jreleaser/jreleaser/commit/a2fed368653ffd6be8930eb6e5beec92a3cbf858

PR #836, opened 2022-05-30
Fix cosign-support by verifying against err-output

### LINKED ISSUE #835: Cosign verify not working
### Steps to Reproduce

Add the following configuration to **jreleaser.yml**

```
signing:
  active: ALWAYS
  enabled: true
  armored: true
  mode: cosign
  cosign:
    version: 1.8.0
    privateKeyFile: cosign.key
    publicKeyFile: cosign.pub
```

Under both Linux (inside a GitHub-action) and on MacOS X (x86) you will see that jreleaser is downloading cosign in the
correct version but fails to verify it.

The reason is that jreleaser validates the version against stdout while cosign is reporting its version via err-out.

This makes cosign currently unusable with jreleaser.

I have already prepared a fix and will provide a PR shortly :)

### Environment Information

- **Operating System**: Linux and OSX
- **JReleaser Version:** 1.0.0
- **JDK Version:** 11 and 17

codepitbull:
Fixes #835

### Context

DownloadableTool should be able to also verify the err-output.
In the concrete case cosign is reporting its version via the error-channel, resulting in verify to always fail.

### Checklist

### DIFF
--- a/sdks/tool-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
+++ b/sdks/tool-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
@@ -32,12 +32,14 @@ public class AbstractTool {
     protected final JReleaserContext context;
     protected final DownloadableTool tool;
     protected final String name;
+    protected final boolean verifyErrorOutput;
 
-    public AbstractTool(JReleaserContext context, String name, String version) {
+    public AbstractTool(JReleaserContext context, String name, String version, boolean verifyErrorOutput) {
         requireNonBlank(version, "'version' must not be blank");
         this.name = requireNonBlank(name, "'name' must not be blank");
         this.context = context;
-        this.tool = new DownloadableTool(context.getLogger(), name, version, PlatformUtils.getCurrentFull());
+        this.verifyErrorOutput = verifyErrorOutput;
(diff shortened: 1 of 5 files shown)


sdks/tool-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
```diff
@@ -32,12 +32,14 @@ public class AbstractTool {
     protected final JReleaserContext context;
     protected final DownloadableTool tool;
     protected final String name;
+    protected final boolean verifyErrorOutput;
 
-    public AbstractTool(JReleaserContext context, String name, String version) {
+    public AbstractTool(JReleaserContext context, String name, String version, boolean verifyErrorOutput) {
```
aalmiray, line 37: Please add an overloaded constructor with the last boolean flag set to `false` as default. This would allow the `Upx.java` tool to remain unchanged.

aalmiray (changes requested):

sdks/tool-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
codepitbull: Done :)

aalmiray:
Thank you for the PR! 😄 

Note that I've squashed your changes to comply with conventional commits as explained at https://github.com/jreleaser/jreleaser/blob/main/CONTRIBUTING.adoc

PR #972, opened 2022-10-03
fix: Ease the regex search for contributors

### LINKED ISSUE #957: [changelog] Contributor exclusion regex does not work
### Steps to Reproduce

Configure a contributor name such as `.*bot` or `*bot`.

### Expected Behaviour

This should exclude contributors such as `liquibot` or `dependabot`.

### Actual Behaviour

This is not the case. Both contributors are still listed.

Only the expression `bot` seems to work.

Feel free to try with https://github.com/liquibase/liquibase-neo4j/ (`mvn jreleaser:changelog -Prelease`).

### Environment Information

- **Operating System**: Mac OS 12.6
- **JReleaser Version:** 1.2.0
- **JDK Version:** Java 17.0.3 / Azul
- **Maven version**: 3.8.6

Sironheart:
Fixes #957?

### Context

https://github.com/jreleaser/jreleaser/issues/957

### Checklist

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/Changelog.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/Changelog.java
@@ -940,7 +940,7 @@ public final class Changelog extends AbstractModelObject<Changelog> implements D
             if (isNotBlank(name)) {
                 String n = name.trim();
                 for (String contributor : contributors) {
-                    if (n.contains(contributor) || n.matches(toSafeRegexPattern(contributor))) {
+                    if (n.contains(contributor) || n.matches(contributor)) {
                         return true;
                     }
                 }
--- a/sdks/jreleaser-git-java-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
+++ b/sdks/jreleaser-git-java-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
@@ -52,7 +52,9 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static java.util.stream.Collectors.toList;
 import static org.mockito.ArgumentMatchers.anyBoolean;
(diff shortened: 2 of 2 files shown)


aalmiray:
hi @Sironheart, thank you for taking the time to send a PR. However, the proposed solution would skip any name that has "bot" on it. Most bots would be skipped (yay) but also any human account with "bot" (booo). I think it's best to not use an implicit value but rather leave it to the user to decide.

The actual problem is that the `*bot` regex is turned into `.*\*bot.*` by the `toSafeRegexPattern()` method. The better fix would be to not invoke said method and use the regex as is. This means the proper regex to skip accounts ending with "bot" would be `.*bot`.

sdks/jreleaser-git-java-sdk/src/test/java/org/jreleaser/sdk/git/ChangelogGeneratorUnitTest.java
```diff
@@ -48,11 +48,7 @@
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
```
aalmiray, line 51: Please leave imports in expanded form. Do not use `*` imports.

Sironheart:
Hey @aalmiray, just to make it clear: You'd like to remove the usage of the regex in this case (i.e. removing the second condition in the if) and would rather check for user defined postfixes? Does this sound right for you? :)

Thanks for your time until this point, would love to continue the contribution ☺️

aalmiray:
@Sironheart close. I still want to use the regex, but not the call to `toSafeRegexPattern()` as that changes the regex itself. The regex should be used "as is" written by the user, there should be no transformation at all.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/Changelog.java
```diff
@@ -940,7 +940,11 @@ public boolean containsContributor(String name) {
             if (isNotBlank(name)) {
                 String n = name.trim();
                 for (String contributor : contributors) {
-                    if (n.contains(contributor) || n.matches(toSafeRegexPattern(contributor))) {
+                    if(n.endsWith("bot")) {
```
aalmiray, line 943: These 3 lines are the ones I'm not so sure about as this behavior is not configurable at the moment and in some cases it might filter a valid account with no way to "unfilter". I think it's better if users to explicitly skip bots for the time being. Exposing a configuration option for this might be a bit messy with what we already have in place.

Thus I think the minimum fix is to remove these 3 lines and keep the change you made in line 947. Thank you 😅

Sironheart: I wasn't quite sure if I am missing a simple solution to receiving the context in this class, thus I initially implemented it this way. Your answer confirms my 'fear' of not being able to receive context/ environment information right now. 

I removed the `endsWith` check. Hopefully you find an easy way to integrate the bot-check in an easy manner :)

aalmiray:
Thank you 😄

PR #1090, opened 2022-12-08
reference closed issues in changelog

### LINKED ISSUE #806: [changelog] reference closed issues
I am evaluating jreleaser as a replacement for semantic-release for changelog generation. jreleaser doesn't mention the closed issues that are referenced in a conventional commit.

for example, the following commit message:

```
feat: add FileHeader.getFileName

which returns the Unicode filename if it exists, and the ASCII filename otherwise
the existing methods FileHeader.getFileNameString and FileHeader.getFileNameW are deprecated

closes #34, closes #53
```

produced the following release note: https://github.com/junrar/junrar/releases/tag/v7.2.0

It would be nice if jreleaser could parse the references to issues within commit messages, and add them to the changelog.

In a conventional commit, an issue could be referenced in the footer, like shown above, or directly in the body, for example:

```
feat: a new feature

this continues the work done in #3 and lays the foundation for #5
```

**Describe alternatives you've considered**
other tools like semantic-release can do this

gotson:
Fixes #806

I reused the `ChangelogProvider.extractIssues` method so that what's displayed in the changelog is the same as the issues that are commented upon to advise they have been released.

I have placed the logic inside the `Commit` class, since it can apply to any type of commit, not just conventional ones.
I could not find a Mustache way to display something once if a list is not empty, so i added a `commitIssuesCount` property. If you know how to do so, we could get rid of that extra property, however everything i tried didn't work.

Couple of things you may want to double-check:
- i have modified the default format of the conventional-commits preset
- i have **not** modified the default format
- since there is an info log in `extractIssues`, it is logged once for each commit, and may confuse the user as to why the same log line is repeated many times
- the issues in the Mustache context are always prefixed by `#`. This is done so that pre-computed links look better, and to ensure links that would be rendered by Github or others would always be rendered as links, so the `#` doesn't need to be added in the template. It may be a bit restrictive, but i think that should be fine with most users.

The PR is separated in multiple commits, as some are fixes that are required but not directly related to the new feature. Feel free to keep as-is, or squash when merging. I thought you would appreciate the split, even just to see them separately.

### Checklist

### DIFF
--- a/sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
+++ b/sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
@@ -53,6 +53,7 @@ import java.util.stream.IntStream;
 import java.util.stream.StreamSupport;
 
 import static java.lang.System.lineSeparator;
+import static java.util.Collections.singletonMap;
 import static java.util.Collections.unmodifiableList;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.joining;
@@ -353,6 +354,7 @@ public class ChangelogGenerator {
         commits.stream()
             .sorted(revCommitComparator)
             .map(rc -> "conventional-commits".equals(changelog.getPreset()) ? ConventionalCommit.of(rc) : Commit.of(rc))
+            .map(c -> c.extractIssues(context))
             .peek(c -> {
                 if (!changelog.getContributors().isEnabled()) return;
 
@@ -371,6 +373,7 @@ public class ChangelogGenerator {
 
         BaseReleaser releaser = context.getModel().getRelease().getReleaser();
         String commitsUrl = releaser.getResolvedCommitUrl(context.getModel());
+        String issueTracker = releaser.getResolvedIssueTrackerUrl(context.getModel(), true);
 
         Map<String, Object> props = context.fullProps();
         StringBuilder changes = new StringBuilder();
@@ -385,7 +388,7 @@ public class ChangelogGenerator {
             final String categoryFormat = resolveCommitFormat(changelog, category);
 
             changes.append(categories.get(categoryKey).stream()
-                    .map(c -> resolveTemplate(categoryFormat, c.asContext(changelog.isLinks(), commitsUrl)))
+                    .map(c -> resolveTemplate(categoryFormat, c.asContext(changelog.isLinks(), commitsUrl, issueTracker)))
                     .collect(joining(lineSeparator)))
                 .append(lineSeparator)
                 .append(lineSeparator());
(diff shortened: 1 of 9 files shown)


sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
```diff
@@ -608,6 +612,11 @@ Map<String, Object> asContext(boolean links, String commitsUrl) {
             context.put("commitTitle", passThrough(title));
             context.put("commitAuthor", passThrough(author.name));
             context.put("commitBody", passThrough(body));
+            context.put("commitIssuesCount", issues.size() > 0 ? issues.size() : null);
```
aalmiray, line 615: I've used a boolean in the past to  mark that a collection may be empty. Setting `null` for a count might not be a good idea. Setting `0` would better unless Mustache's thruthy value takes `0` as true.

sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
```diff
@@ -608,6 +612,11 @@ Map<String, Object> asContext(boolean links, String commitsUrl) {
...
+            context.put("commitIssues", issues.stream().map(i -> {
+                String issue = links ? passThrough("[#" + i + "](" + issueTrackerUrl + i + ")") : "#" + i;
+                return Collections.singletonMap("issue", issue);
```
aalmiray, line 618: Prefer the use of static imports

sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
gotson: I tried 0 and it doesn't work. I can use a boolean if you prefer.

sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
gotson: would that be for `Collections.singletonMap` ?

aalmiray: Yes, sorry. I should had been more explicit.

sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java
aalmiray: Yes, I think the boolean will make it consistent. The name should be `commitHasIssues`.

gotson:
I found, and fixed another issue with extractIssues, where all the `#5` would be matched, even if it was preceded by the repo and it would not match.

`Closes: someone/repo#1, owner/repo#2, repo#3, (#4)` previously all 4 issues would be returned, now if the canonical repo name is`owner/repo`, only `2` and `4` are matched.

aalmiray:
Indeed, we would need the additional form as well. Updating the regex is the way to go.

gotson:
> Indeed, we would need the additional form as well. Updating the regex is the way to go.

Done!

aalmiray (approved):

aalmiray:
Thank you! 😄

PR #1113, opened 2022-12-17
fix: github user lookup strategy

### LINKED ISSUE #1112: [changelog] Contributor username lookup
Discussed in #1111 

### Task List

### Steps to Reproduce

Find a user who -
1. Doesn't have a public email On Github profile
2. Display name (eg. Manik Magar) has more than one profile on Github.

If that user has a commit on the repo, observe his entry in the change log.

### Expected Behaviour

When generating the changelog, the contributor's correct username is added.

### Actual Behaviour

Looks like it is searching by Author's name - 'Kevin King' (see web query [here](https://github.com/search?l=&p=7&q=fullname%3AKevin+King+type%3Auser&type=Users)) and just picked up the first user found
https://github.com/jreleaser/jreleaser/blob/main/sdks/jreleaser-github-java-sdk/src/main/java/org/jreleaser/sdk/github/XGithub.java#L103-L111 -

But that's not the Kevin who committed the change :). This user also has [a private github email](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#:~:text=GitHub%20account.%22-,Note%3A,-If%20you%20created) id - [ID+USERNAME@users.noreply.github.com](mailto:ID+USERNAME@users.noreply.github.com). The API documentation states that users having "Public email" listed on their profile can only be searched using email and since there is no public email listed here, search by email id is not useful. Searching by names as seen below can result in unexpected results.

### From the discussion -

Lookup should - 
> 1. check email, if private then extract username. done.
> 2. search by email. If found use name.
> 3. use name as extracted from commit message (this would need to be recorded as well. I think only the email is recorded at the moment.

manikmagar:
Fixes #1112 

### Context

See Discussion #1111 Changelog picking up wrong username

This PR uses the private email id, if present, to fetch github user profile.

### Checklist


manikmagar:
RE https://github.com/jreleaser/jreleaser/issues/1112#issuecomment-1353618559 -
image

---

I probably don't see that happening. The modified code would not result in partially filled `User` instance. The calling code will still behave the same if returned result is empty (as when even the search by name did not find anything).

I think, `XGithub.findUser` being a lookup method, should not return a partial user if it can't fetch one from remote. 

Commit is already extracting Co-author as a contributor and adding it as a committer - https://github.com/jreleaser/jreleaser/blob/main/sdks/jreleaser-git-java-sdk/src/main/java/org/jreleaser/sdk/git/ChangelogGenerator.java#L596-L601. These committers go through same lookup and should be have the same way.

aalmiray:
Thank you! 😄

PR #1269, opened 2023-03-13
fix: default to unversioned jdk dependency in brew (#1224)

### LINKED ISSUE #1224: [package] Homebrew formula should default to "openjdk" dependency (without a version) for Java projects
Current default behaviour of JReleaser's brew packager is to add a dependency to "openjdk@VERSION".

This should be changed to "openjdk" without a version:

- Most Java software does not need to be pinned to a specific JDK release, it just has a _minimum_ requirement (e.g. 8 or 11).
- Adding "openjdk" as a dependency will ensure that the _latest_ available OpenJDK release is installed (e.g. 19.0.2 as of this writing).
- Using a specific JDK version for each project may clutter the machine with several JDKs, which is especially useless if those aren't LTS versions.
- In those cases where a specific JDK version must be used, one can still add a specific version dependency (or use a bundled JDK in the first place).

This is also the pattern used by most Homebrew formulae for Java software: 

> Most Java-based packages in Homebrew use the default Homebrew-provided openjdk package for their JDK dependency, although some packages with specific version requirements may use a versioned package such as openjdk@8.
https://docs.brew.sh/Homebrew-and-Java#executing-commands-using-a-different-jdk

klausbrunner:
Fixes jreleaser/jreleaser#1224

### Context
See jreleaser/jreleaser#1224 . Automatic JRE/JDK ("openjdk" or "java") dependency without a version is only added if there isn't one already.

### Checklist

### DIFF
--- a/core/jreleaser-engine/src/main/java/org/jreleaser/packagers/BrewPackagerProcessor.java
+++ b/core/jreleaser-engine/src/main/java/org/jreleaser/packagers/BrewPackagerProcessor.java
@@ -32,6 +32,7 @@ import org.jreleaser.util.PlatformUtils;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -138,6 +139,9 @@ BrewPackagerProcessor extends AbstractRepositoryPackagerProcessor<BrewPackager> 
     private static final String FORMULA_MULTI_RB = "formula-multi.rb";
     private static final String RB = ".rb";
     private static final String SKIP_JAVA = "skipJava";
+    private static final String USE_VERSIONED_JAVA = "useVersionedJava";
+
+    private static final List<String> BREW_JDK_ALIASES = Arrays.asList("openjdk", "java");
 
     public BrewPackagerProcessor(JReleaserContext context) {
         super(context);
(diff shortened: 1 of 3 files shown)


aalmiray:
Thanks for the PR! 😄 

However this change constitutes a breaking change as now one would have to explicitly request `openjdk@11` for example. Given this, it's better to make this a 2 step process:
1. deprecate & warn but keep current behavior
2. use new behavior in `v2`

As such, this PR cannot be merged as it is right now. Could you rework the code to keep the old behavior but also incorporate your enhancement? Thanks.

aalmiray:
This `useDefaultJDK` property would have to be assumed to be `true` by default even if not set. We turn it off and assume `false` if unset in `2.x`.

core/jreleaser-engine/src/main/java/org/jreleaser/packagers/BrewPackagerProcessor.java
```diff
@@ -138,6 +138,9 @@
     private static final String FORMULA_MULTI_RB = "formula-multi.rb";
     private static final String RB = ".rb";
     private static final String SKIP_JAVA = "skipJava";
+    private static final String USE_VERSIONED_JAVA = "useVersionedJava";
+
+    private static final List<String> BREW_JDK_ALIASES = List.of("openjdk", "java");
```
aalmiray, line 143: Can't use this API call because codebase mist remain Java8 compatible. Use `Collections.singleton()` instead.

aalmiray:
Build fails because PRs don't have access to organization variables.

core/jreleaser-engine/src/main/java/org/jreleaser/packagers/BrewPackagerProcessor.java
```diff
@@ -278,6 +281,17 @@ protected void fillPackagerProperties(TemplateContext props, Distribution distri
...
     }
 
+    private boolean shouldAddJavaDependency(Distribution distribution) {
+        if ((distribution.getType() == org.jreleaser.model.Distribution.DistributionType.JAVA_BINARY ||
+                distribution.getType() == org.jreleaser.model.Distribution.DistributionType.SINGLE_JAR) &&
+                !isTrue(packager.getExtraProperties().get(SKIP_JAVA))) {
+            return packager.getDependenciesAsList().stream()
+                    .map(BrewPackager.Dependency::getKey).noneMatch(BREW_JDK_ALIASES::contains);
+        } else {
+            return false;
```
aalmiray, line 291: Could use return statement without `else`.

aalmiray:
CodeQL workflow failed for the same reason above: no access to repo/org variables. Will have to update it as well.

klausbrunner:
doc changes in jreleaser/jreleaser.github.io/pull/53

aalmiray:
LGTM. Thank you! 😄

Will be on the road the next couple of days. Will likely merge by the weekend barring any unforeseen issues.

klausbrunner:
Updated the logic to catch pre-defined JDK dependencies both with or without versions set. And promptly ran into #1273 while testing 🙂

aalmiray:
Thank you 😄

PR #1277, opened 2023-03-16
Draft: Adding FeignLogger that wraps JReleaserLogger

bdemers:
Sets the log level to full (wire logging), but will only output if JReleaser logger debug level is enabled

Fixes #?

### Context

I was trying to figure out why GitHub username was not getting resolved but didn't know what the API was returning. I had to add a feign logger to do so.

### Checklist

### DIFF
--- a/sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/ClientUtils.java
+++ b/sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/ClientUtils.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import feign.Client;
 import feign.Feign;
+import feign.Logger;
 import feign.Request;
 import feign.form.FormData;
 import feign.form.FormEncoder;
@@ -36,6 +37,7 @@ import org.jreleaser.model.JReleaserVersion;
 import org.jreleaser.model.internal.JReleaserModelPrinter;
 import org.jreleaser.model.spi.announce.AnnounceException;
 import org.jreleaser.model.spi.upload.UploadException;
+import org.jreleaser.sdk.commons.feign.FeignLogger;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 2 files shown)


sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/feign/FeignLogger.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private final JReleaserLogger logger;
+
+    public FeignLogger(JReleaserLogger logger) {
+        this.logger = logger;
+    }
+
+    @Override
+    protected void log(String configKey, String format, Object... args) {
+        logger.debug(String.format(methodTag(configKey) + format, args));
```
bdemers, line 32: This logic is _copied_ from the Feign SLF4J impl, they use String.format directly, because the template string may not match the target impl's logger (e.g. SLF4F's template string)

JReleaserLogger doesn't have an `ifDebugLevel()` method, so this means the string will be formatted even if the debug logger is disabled.

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/ClientUtils.java
```diff
@@ -107,6 +109,8 @@ public static Feign.Builder builder(JReleaserLogger logger,
         }
 
         return builder
+            .logger(new FeignLogger(logger))
+            .logLevel(Logger.Level.FULL)
```
bdemers, line 113: First time using Feign directly, but this doesn't seem ideal, I'm not sure if this property should map back to the JReleaser log level (trace -> full, debug -> basic, etc). But that info isn't available in the JReleaserLogger interface, so I just hard-coded it to `FULL` 🤷

bdemers:
@aalmiray feel free to just delete this, I figured I'd pass it along because I had it handy.
The root cause of my issue was my commits use a secondary email, which do not appear to be returned by a user search:
`https://api.github.com/search/users?q={email}`

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/feign/FeignLogger.java
aalmiray: May need to add level guards to `JReleaserLogger` to avoid unused formatting.

aalmiray: Switching to JSL's `MessageFormat` instead of SLF4J may be an option as well.

aalmiray:
Logging in Feign may be too chatty. I wonder how it could be activated in such way that it/s not to the detriment of the tool. You definitely want to have as much information when something goes wrong, OTOH you don't want to pay upfront for something that's not needed.

aalmiray:
Closing in favor of https://github.com/jreleaser/jreleaser/issues/1638

PR #1327, opened 2023-04-15
fix: Add validation for tags to ChocolateyPackagerValidator

### LINKED ISSUE #1313: [packagers] Chocolatey requires project tags to be non empty but jreleaser is not checking that
The `<tags></tags>` entry in the nuspec file must contain at least one tag to publish it to the repository, so jreleaser should maybe warn if that is missing.

SimonVerhoeven:
Fixes #1313

### Context

Adds validation for the tag list not being empty

### Checklist

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/packagers/ChocolateyPackagerValidator.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/packagers/ChocolateyPackagerValidator.java
@@ -250,6 +250,10 @@ public final class ChocolateyPackagerValidator {
 
         Project project = context.getModel().getProject();
 
+        if (distribution.getTags().isEmpty()) {
+            errors.configuration(RB.$("validation_must_not_be_empty", "distribution.tags"));
+        }
+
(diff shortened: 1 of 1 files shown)


core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/packagers/ChocolateyPackagerValidator.java
```diff
@@ -250,6 +250,10 @@ public static void postValidateChocolatey(JReleaserContext context, Distribution
 
         Project project = context.getModel().getProject();
 
+        if (distribution.getTags().isEmpty()) {
+            errors.configuration(RB.$("validation_must_not_be_empty", "tags"));
```
aalmiray, line 254: Should probably be more specific with the missing property name. Please update it to "distribution.tags". Thanks.

SimonVerhoeven: Of course, I was a tad on the fence between the two myself.

aalmiray:
Thanks! 😄

Issue #1357, opened 2023-04-23
[release] Show the reason when RestAPIException fails a build


ge0ffrey:
See also https://github.com/jreleaser/jreleaser/discussions/1356

aalmiray:
🎉 This issue has been resolved in `v1.6.0` ([Release Notes](https://github.com/jreleaser/jreleaser/releases/tag/v1.6.0))

PR #1358, opened 2023-04-23
fix: Show the reason when RestAPIException fails a build

ge0ffrey:
Fixes https://github.com/jreleaser/jreleaser/issues/1357


sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/RestAPIException.java
```diff
@@ -40,6 +40,7 @@ public RestAPIException(int status, String reason) {
     }
 
     public RestAPIException(Request request, int status, String reason, Map<String, Collection<String>> headers) {
+        super(status + ": " + reason);
```
ge0ffrey, line 43: I am not sure if reason doesn't already include the status code too.

FWIW, it wasn't practical to verify that. So you might see the status code twice, which is still much better than not seeing it at all.

PR #1389, opened 2023-05-17
feat(release): support bitbucket cloud releases

### LINKED ISSUE #12: Support releasing to BitBucket

viqueen:
Fixes #12 

### Context
- supports releasing to bitbucket cloud repository
- it does **not** include resolving an issue tracker (yet) , to be continued
   - we can support default issue tracker first with [bitbucket cloud group issue tracker api](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-issue-tracker/) 
   - and explore integration with jira for [version/release management}(https://support.atlassian.com/jira-software-cloud/docs/enable-releases-and-versions/)
- it includes artifact upload to bitbucket cloud
- **note**: 
    - I ran into some issues with git push using ssh locally, other than that everything works fine with repo access tokens and https remotes
    

### Checklist

- [ ] register BitbucketcloudUploader with JsonSchemaGenerator
- [ ] use `Templates.resolveTemplate()` where applicable
- [ ] use `bitbucket` dsl with `mode` set to CLOUD vs DC/SERVER

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/BitbucketcloudReleaser.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/BitbucketcloudReleaser.java
@@ -1,0 +1,284 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Copyright 2020-2023 The JReleaser authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jreleaser.model.internal.release;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.jreleaser.model.Active;
+import org.jreleaser.model.api.common.CommitAuthor;
+import org.jreleaser.model.api.release.Changelog;
+
+import java.util.Map;
+
+import static java.util.Collections.unmodifiableMap;
+
+/**
+ * @author Hasnae Rehioui
+ * @since 1.7.0
+ */
+public class BitbucketcloudReleaser extends BaseReleaser<org.jreleaser.model.api.release.BitbucketcloudReleaser, BitbucketcloudReleaser> {
+    private static final long serialVersionUID = 3160208925103963195L;
+
+    @JsonIgnore
+    private final org.jreleaser.model.api.release.BitbucketcloudReleaser immutable = new org.jreleaser.model.api.release.BitbucketcloudReleaser() {
+        private static final long serialVersionUID = -7158952152230034613L;
+
+        @Override
+        public String getServiceName() {
+            return BitbucketcloudReleaser.this.getServiceName();
+        }
+
+        @Override
+        public boolean isReleaseSupported() {
+            return BitbucketcloudReleaser.this.isReleaseSupported();
+        }
+
+        @Override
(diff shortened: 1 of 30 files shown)


aalmiray:
Is there a significant API difference between BBC and BBS? I'd suppose some would like to use their own BBS instance when running on premise.

viqueen:
> Is there a significant API difference between BBC and BBS? I'd suppose some would like to use their own BBS instance when running on premise.

Historically they are different products, so there's definitely a difference in the API endpoints, but also in access token management.

aalmiray:
Would it make sense to support both with a single DSL element named `bitbucket` and additional enum property `mode` with CLOUD and SERVER as value?

The reason I ask is that both GitHub and Gitlab support managed and self hosted services with a single DSL element because the only difference at the moment is the host's value, given that the API is the same for both scenarios.

aalmiray:
I'm thinking that support for BBS may be added later based on user request. Both BBC & BBS could be supported by a single DSL element as previously suggested, in which case `bitbucketcloud` would be deprecated in favor of `bitbucket`.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/BitbucketcloudReleaser.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+    };
+
+    public BitbucketcloudReleaser() {
+        super(org.jreleaser.model.api.release.BitbucketcloudReleaser.TYPE, true);
+        setHost("bitbucket.org");
+        setRepoUrl("https://{{repoHost}}/{{repoOwner}}/{{repoName}}");
+        setRepoCloneUrl("https://{{repoHost}}/{{repoOwner}}/{{repoName}}.git");
+        setCommitUrl("https:://{{repoHost}}/{{repoOwner}}/{{repoName}}/commits");
+        setSrcUrl("https://{{repoHost}}/{{repoOwner}}/{{repoName}}/src/{{repoBranch}}");
```
aalmiray, line 272: I think there are other places that expect a non-null value for issueTracker like `project.links` and the changelog generator. Will have to null proof them after merging.

sdks/jreleaser-bitbucketcloud-java-sdk/src/main/java/org/jreleaser/sdk/bitbucketcloud/BitbucketcloudArtifactUploader.java
```diff
@@ -0,0 +1,87 @@
...
+        for (Artifact artifact : artifacts) {
+            Path path = artifact.getEffectivePath(context);
+            context.getLogger().info(" - {}", path.getFileName());
+
+            if (context.isDryrun()) {
+                continue;
+            }
+
+            try {
+                Response response = api.uploadArtifact(uploader.getProjectIdentifier(), uploader.getPackageName(), path);
```
aalmiray, line 78: project identifier might contain name templates which should be resolved before calling the api. Use `Templates.resolveTemplate()` to get the resolved value.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/upload/BitbucketcloudUploader.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.Map;
+
+import static java.util.Collections.unmodifiableMap;
+import static org.jreleaser.mustache.Templates.resolveTemplate;
+
+/**
+ * @author Hasnae Rehioui
+ * @since 1.7.0
+ */
+public class BitbucketcloudUploader extends AbstractGitPackageUploader<org.jreleaser.model.api.upload.BitbucketcloudUploader, BitbucketcloudUploader> {
```
aalmiray, line 33: This uploader should be registered with JsonSchemaGenerator in the same way as others uploaders have been registered.

aalmiray (changes requested):

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/BitbucketcloudReleaser.java
aalmiray: Did a quick check and this won't be a problem.

PR #1497, opened 2023-09-29
fix(assemble): Add missing template option bindings to jlink

### LINKED ISSUE #1496: [assemble] JLink assembler launcher missing main jar
### Task List

- [ ] Steps to reproduce provided

### Steps to Reproduce

When jlink is assembled a template variable `distributionJavaMainJar` in https://github.com/jreleaser/jreleaser/blob/e4a487d85d5901bf2b46b45cc46cbcf8f512dd33/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/jlink/jlink/bin/launcher.tpl#L142 is empty, thus it creates a non-working launcher script. Same type of launcher script in `JavaArchive` works fine what comes for main jar.

I was looking differences between `JlinkAssemblerProcessor` and `JavaArchiveAssemblerProcessor` and latter seems to set properties in https://github.com/jreleaser/jreleaser/blob/e4a487d85d5901bf2b46b45cc46cbcf8f512dd33/core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JavaArchiveAssemblerProcessor.java#L97 while former don't do no customisation for template properties.

I compiled 1.9.0 snapshot locally and modified `JlinkAssemblerProcessor` to set those properties similarly to `JavaArchiveAssemblerProcessor` and it did fix the issue. 

### Expected Behaviour

JLink should have working launcher script with fatjar.

Something like:
```
set -- \
        -classpath "$CLASSPATH" \
        -jar "${JARSDIR}/spring-cli-0.7.1.jar" \
        "$@"
```

### Actual Behaviour

JLink launcher is missing path to an actual jar file.

```
set -- \
        -classpath "$CLASSPATH" \
        -jar "${JARSDIR}/" \
        "$@"
```

This is for spring-cli where I'm trying to add "standalone" release similarly to JReleaser itself. Config file is https://github.com/spring-projects-experimental/spring-cli/blob/main/jreleaser.yml

jvalkeal:
Fix missing template variable binding in jlink assembler.

Add distributionJavaMainModule, distributionJavaMainClass and distributionJavaMainJar to template context in JlinkAssemblerProcessor similarly it's done in JavaArchiveAssemblerProcessor.

This fix don't try to add distributionJavaOptions as that's not used in a jlink template and assembler structure seems to be slightly different compared to java-archive.

Fixes #1496

I didn't find any tests to tweak.

### DIFF
--- a/core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JlinkAssemblerProcessor.java
+++ b/core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JlinkAssemblerProcessor.java
@@ -48,6 +48,9 @@ import static org.jreleaser.assemblers.AssemblerUtils.readJavaVersion;
 import static org.jreleaser.model.Constants.KEY_ARCHIVE_FORMAT;
 import static org.jreleaser.model.Constants.KEY_DISTRIBUTION_ASSEMBLE_DIRECTORY;
 import static org.jreleaser.model.Constants.KEY_DISTRIBUTION_EXECUTABLE;
+import static org.jreleaser.model.Constants.KEY_DISTRIBUTION_JAVA_MAIN_CLASS;
+import static org.jreleaser.model.Constants.KEY_DISTRIBUTION_JAVA_MAIN_JAR;
+import static org.jreleaser.model.Constants.KEY_DISTRIBUTION_JAVA_MAIN_MODULE;
 import static org.jreleaser.mustache.Templates.resolveTemplate;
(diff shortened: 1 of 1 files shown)


jvalkeal:
What I'm kinda questioning if these changes should go to`AbstractJavaAssembler.props()`.

aalmiray:
Thanks. I'll think it's OK to keep the changes where they are at the moment. There are a couple of updates planned for `2.x` that will removed duplication like this but requires breaking compatibility.

aalmiray:
Thank you! 😄

PR #1727, opened 2024-09-19
build: Cancel concurrent PR jobs

gastaldi:
GitHub will automatically cancel concurrent jobs per PRs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ permissions:
   contents: read
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}
```
aalmiray, line 10: Is `matrix.os` resolvable here? Or should it be the runner's OS?

gastaldi: https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs says it is

aalmiray:
Thanks

aalmiray:
@gastaldi as I thought, the `matrix` element is not implicitly defined https://github.com/jreleaser/jreleaser/actions/runs/10971972024

This means PR's won't run at all 😢

gastaldi:
Great, so the docs is misleading in saying that `matrix` is allowed. Anyway, it should work if you remove just that expression, let me submit a PR

PR #1864, opened 2025-04-30
feat(deploy): Add Maven metadata transformation for snapshot releases

marko-bekhta:
This one is from the discussion at https://github.com/jreleaser/jreleaser/discussions/1862

I'll open it as a draft as it only adds this to nexus2 for now, and I wanted to get some early feedback if I'm going in the right direction here, as I'm unfamiliar with the code 🙈 😃 (will create a GH issue once this is in a more ready shape)

- I see that not all maven deployers have the snapshot URLs, and I'm also not familiar with them, so my thinking here was that if we add the "merge" logic to the `AbstractMavenDeployer` we could use the optional URL for the metadata, and those actual deployers that don't support the feature would always return the `DISABLED` transformation mode + empty optional links as a second precaution. 
- I've tried to keep most of the changes to the spi/internal packages. 
- Not sure if it is ok to add the `dev.failsafe:failsafe` to `jreleaser-java-sdk-commons.gradle`. I initially wanted to have the download logic inside the specific deployers (where the dependency is already present), but I got lost a bit (how to cleanly pass the downloaded XML back to the `AbstractMavenDeployer`), and hence, this is where I ended up so far. 
- I've hardcoded the retry "policy" for the XML downloads as I wasn't sure if that's something that would be much valuable for the user to configure. 
- There are no docs updates for now, but once I figure this out, I'll see what/where to add to the docs as well.

### Checklist

### DIFF
--- a/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/deploy/maven/Nexus2MavenDeployer.java
+++ b/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/deploy/maven/Nexus2MavenDeployer.java
@@ -32,6 +32,8 @@ public interface Nexus2MavenDeployer extends MavenDeployer {
 
     String getSnapshotUrl();
 
+    String getSnapshotServiceUrl();
+
     String getVerifyUrl();
 
     boolean isCloseRepository();
@@ -49,6 +51,8 @@ public interface Nexus2MavenDeployer extends MavenDeployer {
     Stage getStartStage();
 
     Stage getEndStage();
+
+    MavenMetadataTransformationMode getMavenMetadataTransformationMode();
 
     enum Stage {
         UPLOAD,
--- a/api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
+++ b/api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
@@ -127,6 +127,13 @@ ERROR_artifact_does_not_exist             = Artifact does not exist. {}
 ERROR_unexpected_error_hash_read          = Unexpected error when reading hash from {}
 ERROR_unexpected_error_calculate_checksum = Unexpected error calculating checksum for {}
 
+mavenMetadataTransformation.header               = Transforming Maven metadata XMLs
+mavenMetadataTransformation.processing           = Transforming Maven metadata XML in: {}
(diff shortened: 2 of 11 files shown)


aalmiray:
I would rather add the merge behavior to the abstract class without any explicit logic to trigger it using the snapshotUrl. Each deployer would have the choice to trigger it when needed. 

Though, deployers have a flag named `snapshotSupported` and the context can tell you if the project is snapshot or not. Thus, finding out if a merge should occur can happen in the common code. 

Adding failsafe as a dependency to the common subproject should not be an issue.

api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
```diff
@@ -127,6 +127,13 @@ ERROR_artifact_does_not_exist             = Artifact does not exist. {}
 ERROR_unexpected_error_hash_read          = Unexpected error when reading hash from {}
 ERROR_unexpected_error_calculate_checksum = Unexpected error calculating checksum for {}
 
+mavenMetadataTransformation.header               = Transforming Maven metadata XMLs
```
aalmiray, line 130: Please use lowercase and dots instead of CamelCase

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/deploy/maven/Nexus2MavenDeployer.java
```diff
@@ -32,6 +32,8 @@ public interface Nexus2MavenDeployer extends MavenDeployer {
 
     String getSnapshotUrl();
 
+    String getSnapshotServiceUrl();
```
aalmiray, line 35: Why is this needed? Could it be deduced from the snapshot url?

aalmiray (changes requested):

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/deploy/maven/Nexus2MavenDeployer.java
marko-bekhta:
```
https://oss.sonatype.org/content/repositories/snapshots/
https://oss.sonatype.org/service/local/repositories/snapshots/content/
```
they are slightly different, so my thinking here was, just to stay on the safe side, give the user an option to provide the value. Let me try running nexus 2/3 locally and see if the URL structure is predictable and if I can get the repository name (`snapshots`) and reconstruct the other url.

PR #1884, opened 2025-06-01
fix(gradle): add support for blank gpg passphrase

### LINKED ISSUE #1883: [core] signing.passphrase must not be blank.
In a `signing {}` block like so:

```
    jreleaser {
        signing {
            setActive("ALWAYS")
            setMode("COMMAND")
        }
        ...
    }
```

I get the error:

```
[ERROR] signing.passphrase must not be blank.
```

But what if the PGP private key is not protected by a passphrase? It feels like jreleaser should not be enforcing it, should it?

JonasVautherin:
Fixes #1883 

### Context

GPG private keys are not always protected by a passphrase. Jreleaser should support unprotected keys.

### Testing

I tried it both with a passphrase-protected key and an unprotected key, and it works for me.

### Checklist

### DIFF
--- a/api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
+++ b/api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
@@ -160,6 +160,7 @@ signing.file.newer                   = {} is newer than {}
 signing.check.published.key          = Checking if key {} has been published
 signing.key.published                = Key {} was found as published
 signing.key.not.published            = Key {} was NOT found as published
+signing.passphrase.blank             = Passphrase is blank. In case a passphrase is needed, you can configure a value using the {}, or define a System property {}, or define a {} environment variable, or define a key/value pair in {} with a key named {}
 signing.public.key.expiration.date   = Signing key {} expires at {}
 signing.public.key.no.expiration.date = Signing key {} does not expire
 ERROR_signing_verify_file            = Could not verify file {} with signature {}
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
@@ -21,6 +21,7 @@ import org.jreleaser.bundle.RB;
 import org.jreleaser.model.api.JReleaserContext.Mode;
 import org.jreleaser.model.internal.JReleaserContext;
 import org.jreleaser.model.internal.deploy.maven.MavenDeployer;
+import org.jreleaser.model.internal.environment.Environment;
 import org.jreleaser.model.internal.signing.Signing;
 import org.jreleaser.util.DefaultVersions;
 import org.jreleaser.util.Errors;
@@ -38,6 +39,8 @@ import static org.jreleaser.model.api.signing.Signing.GPG_PUBLIC_KEYRING;
 import static org.jreleaser.model.api.signing.Signing.GPG_SECRET_KEY;
 import static org.jreleaser.model.internal.validation.common.Validator.checkProperty;
 import static org.jreleaser.model.internal.validation.common.Validator.resolveActivatable;
+import static org.jreleaser.util.Env.envKey;
+import static org.jreleaser.util.Env.sysKey;
 import static org.jreleaser.util.StringUtils.isBlank;
 
 /**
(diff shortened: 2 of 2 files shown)


core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
```diff
@@ -69,13 +69,19 @@ public static void validateSigning(JReleaserContext context, Mode mode, Errors e
...
-                cosign ? COSIGN_PASSWORD : GPG_PASSPHRASE,
-                "signing.passphrase",
-                signing.getPassphrase(),
-                errors,
-                context.isDryrun()));
+        String passphrase = checkProperty(context,
+            cosign ? COSIGN_PASSWORD : GPG_PASSPHRASE,
+            "signing.passphrase",
+            signing.getPassphrase(),
+            new Errors(),
```
JonasVautherin, line 79: I tried to minimise my changes. I pass `new Errors()` here as a way to ignore the errors (otherwise it just fails).

aalmiray: Yes, passing a new instance of `Errors` is the way to go. You could change the logic to always pass a new instance no matter what. 

Also, the warning should be already resolved from internal message resources. 

Given that the errors instance is new, if the passphrase is blank there should be a single error message. You could grab it and print it as a warning.

JonasVautherin: But if I print the error as a warning, it says this:

```
[WARN]    [validation] signing.passphrase must not be blank. Configure a value using the Gradle DSL, or define a System property jreleaser.gpg.passphrase, or define a JRELEASER_GPG_PASSPHRASE environment variable, or define a key/value pair in /home/jones/.jreleaser/config.properties with a key named JRELEASER_GPG_PASSPHRASE
```

Which works, but it feels a bit weird because `signing.passphrase` can actually be blank, right? This is why I was thinking of logging "[WARN] Passphrase is blank" instead.

aalmiray: Ok, right. Then please add a message resource for the new warning message stating that the passphrase is blank. And if one should be supplies then use the suggested env vars or sys props.

JonasVautherin: Done! What do you think? :blush:

api/jreleaser-resource-bundle/src/main/resources/org/jreleaser/bundle/Messages.properties
```diff
@@ -160,6 +160,7 @@ signing.file.newer                   = {} is newer than {}
 signing.check.published.key          = Checking if key {} has been published
 signing.key.published                = Key {} was found as published
 signing.key.not.published            = Key {} was NOT found as published
+signing.passphrase.blank             = Passphrase is blank. In case a passphrase is needed, you can configure a value using the Gradle DSL, or define a System property {}, or define a {} environment variable, or define a key/value pair in {} with a key named {}
```
aalmiray: The Gradle DSL is not the only one. There should be another message key in the resources file that can be used as a guide.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
```diff
@@ -69,13 +72,25 @@ public static void validateSigning(JReleaserContext context, Mode mode, Errors e
...
+            context.isDryrun());
```
aalmiray: The last parameter can be explicitly set to `false`.

aalmiray (changes requested):

aalmiray (approved):

aalmiray:
Thank you 😄

JonasVautherin:
And thank you for the guidance! And I see that it's already released in early-access :tada:.

Out of curiosity, what is the process to get it to Maven Central? Just to have an idea about when I could expect to see it there?

aalmiray:
Releases to Maven Central are posted every two months, at the end of the month, on even months. The next release is scheduled for June 30th. 

Per our release cadence https://github.com/jreleaser/jreleaser/discussions/873

aalmiray:
🎉 This issue has been resolved in `v1.19.0` ([Release Notes](https://github.com/jreleaser/jreleaser/releases/tag/v1.19.0))

PR #1951, opened 2025-07-28
feat: support for jbang built extensions

### LINKED ISSUE #1952: [extensions] Support loading extensions built with JBang
currently extensions can be either already published GAV's or a directory of jars  - it would be great to use have ability to use jbang scripts/apps which could be built on-demand.

refer to source files
```
extensions:
  github-summary:
    jbang: src/jreleaser/extensions/myext.java
```

```
extensions:
  github-summary:
    jbang: src/jreleaser/extensions/myext.kt
```

and even remote jbang script refs:

```
extensions:
  github-summary:
    jbang: https://gist.github.com/maxandersen/a9486a710dbdea5e65066909171ef4f8
```

maxandersen:
Fixes #1952 

### Context

Adds support for using jbang to define extensions.

```
extensions:
  github-summary:
    jbang: misc/jreleaser-extension/githubjobsummary.java
```

### Checklist

### DIFF
--- a/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
+++ b/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
@@ -34,7 +34,7 @@ public interface Extension extends Domain, EnabledAware {
 
     String getDirectory();
 
-    String getJBang();
+    String getJbang();
 
     List<? extends Provider> getProviders();
 
--- a/core/jreleaser-engine/src/main/java/org/jreleaser/extensions/internal/DefaultExtensionManager.java
+++ b/core/jreleaser-engine/src/main/java/org/jreleaser/extensions/internal/DefaultExtensionManager.java
@@ -422,7 +422,6 @@ public final class DefaultExtensionManager implements ExtensionManager {
             return this;
         }
 
-
         public ExtensionBuilder withExtensionPoint(String type, Map<String, Object> properties) {
             extensionPoints.put(type, new ExtensionPointDef(type, properties));
             return this;
--- a/core/jreleaser-engine/src/main/java/org/jreleaser/extensions/internal/DefaultExtensionManager.java
+++ b/core/jreleaser-engine/src/main/java/org/jreleaser/extensions/internal/DefaultExtensionManager.java
@@ -26,6 +26,7 @@ import org.jreleaser.extensions.api.workflow.WorkflowListener;
 import org.jreleaser.model.JReleaserException;
 import org.jreleaser.model.internal.JReleaserContext;
 import org.jreleaser.sdk.command.CommandException;
+import org.jreleaser.sdk.tool.JBang;
 import org.jreleaser.sdk.tool.Mvn;
 import org.jreleaser.sdk.tool.ToolException;
 import org.jreleaser.templates.TemplateResource;
@@ -129,6 +130,10 @@ public final class DefaultExtensionManager implements ExtensionManager {
 
         if (isNotBlank(extensionDef.getGav())) {
             directory = resolveJARs(context, extensionDef);
+        } 
+
+        if (isNotBlank(extensionDef.getJBang())) {
+            directory = resolveJBangJARs(context, extensionDef);
         }
 
         Path directoryPath = Paths.get(directory);
@@ -230,6 +235,50 @@ public final class DefaultExtensionManager implements ExtensionManager {
         return target.toString();
     }
 
+    private String resolveJBangJARs(JReleaserContext context, ExtensionDef extensionDef) {
+        Path target = context.getOutputDirectory().resolve("extensions")
+            .resolve(extensionDef.getName())
+            .toAbsolutePath();
+
+        JBang jbang = new JBang(context.asImmutable(), DefaultVersions.getInstance().getJBangVersion());
+
+        try {
+            if (!jbang.setup()) {
+                throw new JReleaserException(RB.$("tool_unavailable", "jbang"));
+            }
+        } catch (ToolException e) {
+            throw new JReleaserException(RB.$("tool_unavailable", "jbang"), e);
+        }
+
+        try {
+            FileUtils.deleteFiles(target, true);
+
+            //build and export the extension to a local jar 
+            // jbang export local --force -O out/.../<extname>.jar <scriptref>
+            // this will NOT include the extension dependencies as output
+            // would overlap with jresolver itself.
+            // once jbang supports buildTimeOnly deps we can use that to
+            // get additional deps.
+            String targetJar = target.resolve(extensionDef.getName()).toString();
+            List<String> args = new ArrayList<>();
+            args.add("-q");
+            args.add("export");
+            args.add("local");
+            args.add("--force");
+            args.add("-O");
+            args.add(targetJar);
+            args.add(extensionDef.getJBang());
+
+            context.getLogger().debug(RB.$("extension.manager.jbang.export.jars", extensionDef.getGav(), context.relativizeToBasedir(targetJar)));
+
+            jbang.invoke(context.getBasedir(), args);
+        } catch (IOException | CommandException e) {
+            throw new JReleaserException(RB.$("ERROR_unexpected_error"), e);
+        }
+
+        return target.toString();
+    }
+
     private void processExtension(JReleaserContext context, Extension extension, Set<String> visitedExtensionNames, Set<String> visitedExtensionTypes) {
         String extensionName = extension.getName();
         String extensionType = extension.getClass().getName();
@@ -277,13 +326,15 @@ public final class DefaultExtensionManager implements ExtensionManager {
         private final String name;
         private final String gav;
         private final String directory;
+        private final String jbang;
         private final boolean enabled;
         private final Map<String, ExtensionPointDef> extensionPoints = new LinkedHashMap<>();
 
-        private ExtensionDef(String name, String directory, String gav, boolean enabled, Map<String, ExtensionPointDef> extensionPoints) {
+        private ExtensionDef(String name, String directory, String gav, String jbang, boolean enabled, Map<String, ExtensionPointDef> extensionPoints) {
             this.name = name;
             this.gav = gav;
             this.directory = directory;
+            this.jbang = jbang;
             this.enabled = enabled;
             this.extensionPoints.putAll(extensionPoints);
         }
@@ -300,6 +351,9 @@ public final class DefaultExtensionManager implements ExtensionManager {
             return directory;
         }
 
+        public String getJBang() {
+            return jbang;
+        }
         private boolean isEnabled() {
             return enabled;
         }
@@ -333,6 +387,7 @@ public final class DefaultExtensionManager implements ExtensionManager {
         private final DefaultExtensionManager defaultExtensionManager;
         private String gav;
         private String directory;
+        private String jbang;
         private boolean enabled;
 
         public ExtensionBuilder(String name, DefaultExtensionManager defaultExtensionManager) {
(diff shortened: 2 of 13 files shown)


api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
```diff
@@ -34,6 +34,8 @@ public interface Extension extends Domain, EnabledAware {
 
     String getDirectory();
 
+    String getJBang();
```
aalmiray: Gradle DSL requires a similar update https://github.com/jreleaser/jreleaser/blob/522d6f9fd73600f07cd1fa54798436ecdc60b287/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/dsl/extensions/Extension.groovy#L32

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/extensions/Extension.java
```diff
@@ -43,6 +43,7 @@ public final class Extension extends AbstractModelObject<Extension> implements D
     private String name;
     private String gav;
     private String directory;
+    private String jbang;
```
aalmiray, line 46: Gradle DSL requires a similar update https://github.com/jreleaser/jreleaser/blob/522d6f9fd73600f07cd1fa54798436ecdc60b287/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/dsl/extensions/ExtensionImpl.groovy#L41

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/extensions/Extension.java
```diff
@@ -65,6 +66,11 @@ public String getDirectory() {
             return directory;
         }
 
+        @Override
```
aalmiray, line 69: Please update the serialVersionUID at 

https://github.com/jreleaser/jreleaser/blob/522d6f9fd73600f07cd1fa54798436ecdc60b287/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/extensions/Extension.java#L39
https://github.com/jreleaser/jreleaser/blob/522d6f9fd73600f07cd1fa54798436ecdc60b287/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/extensions/Extension.java#L49

sdks/jreleaser-tool-java-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
```diff
@@ -20,6 +20,7 @@
 import org.jreleaser.bundle.RB;
 import org.jreleaser.model.api.JReleaserContext;
 import org.jreleaser.sdk.command.Command;
+import org.jreleaser.sdk.command.Command.Result;
```
aalmiray: This import seems to be extra. Please run `./gradlew check` to find unnecessary imports

sdks/jreleaser-tool-java-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java
```diff
@@ -88,6 +89,7 @@ public void invoke(Path parent, List<String> args) throws CommandException {
         }
     }
 
+
```
aalmiray: This new line may be removed

sdks/jreleaser-tool-java-sdk/src/main/resources/META-INF/jreleaser/tools/jbang.properties
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+unpack=true
+#should really use "latest"
```
aalmiray, line 20: Would love to use latest but then version matching has to change

core/jreleaser-engine/src/main/java/org/jreleaser/extensions/internal/DefaultExtensionManager.java
```diff
@@ -357,19 +412,25 @@ public ExtensionBuilder withDirectory(String directory) {
...
+            this.jbang = jbang;
+            return this;
+        }
+
         public ExtensionBuilder withEnabled(boolean enabled) {
             this.enabled = enabled;
             return this;
         }
 
+
```
aalmiray: This empty line may be removed

aalmiray (changes requested):

aalmiray:
Build failed due to checkstyle

```
 Error: eckstyle] [ERROR] /home/runner/work/jreleaser/jreleaser/sdks/jreleaser-tool-java-sdk/src/main/java/org/jreleaser/sdk/tool/AbstractTool.java:23:8: Unused import - org.jreleaser.sdk.command.Command.Result. [UnusedImports]
```

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
maxandersen: gotcha - i'm wondering if it would make sense to make room for jbang arguments ? 

i.e. technically you can affect/modify the build with  jbang arguments like `jbang -Dkey=value --compile-options etc.`

maybe that should be a `jbang-args` flag ? 

```
jbang: 
   - scriptref: myext.java
   - args: ...
 ``` 
 
 is an alternative but would be sad to loose the simplicity of `jbang: myext.java` ...or do you allow maps|string option in jreleaser elsewhere?

File: sdks/jreleaser-tool-java-sdk/src/main/resources/META-INF/jreleaser/tools/jbang.properties (hunk 6 above)
Comment (maxandersen) [reply sdks/jreleaser-tool-java-sdk/src/main/resources/META-INF/jreleaser/tools/jbang.properties:20]: having download be for specific version seems fine - but that it seem to want to download older version when newer available is problematic...anyhow - for now i suggest we just keep this and improve tool version handling separately.

File: api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java (hunk 1 above)
Comment (aalmiray) [reply api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java]: Well, there's a map in environment: for extra properties but it may not necessarily be available at the time extensions are loaded. 

Also would prefer if the additional args were set closer to where they are needed.

Comment (aalmiray) [reply api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java]: We would have to decide how to do it as changing the type after the next release would constitute a breaking change. 

What other elements may be required besides `script` and `args`?

File: plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/dsl/extensions/Extension.groovy [hunk 8]
```diff
@@ -36,6 +36,8 @@ interface Extension {
 
     DirectoryProperty getDirectory()
 
+    Property<String> getJbang()
```
maxandersen, line 39: this should be `getJBang()` afaics but if i do that it complains about property naming...

problem or can ignore?

aalmiray: Oh, it's because of the case of the B letter. It should be lower case b

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
maxandersen: similar as https://github.com/jbangdev/jbang-gradle-plugin and maven ones...

the main ones are script, jbang-args and trust (although trust we could consider auto add/remove in calls but feels risky ;)

aalmiray: Ok, so we need 4 properties:

- script
- args
- jbangArgs
- trusts


To keep compatibility with names found in other UIs. This means adding a pair of classes in model-api and model-impl following the patterns for other Domain subtypes. The gradle DSL would require similar changes, which I can take care of.

maxandersen: args itself does not make sense here - as jreleaser does its own init and skips any main method calling.

maxandersen:
I  pushed update that adds recursive scanning of .jar in extension directories + use jbang export portable instead of jbang export local. 

And it works...BUT you get:

out/jreleaser/extensions/github-summary/
├── github-summary.jar
└── lib
    ├── autolink-0.10.0.jar
    ├── bcpg-jdk18on-1.80.jar
    ├── bcprov-jdk18on-1.80.jar
    ├── commonmark-0.21.0.jar
    ├── commonmark-ext-autolink-0.21.0.jar
    ├── commons-codec-1.18.0.jar
    ├── commons-compress-1.27.1.jar
    ├── commons-io-2.16.1.jar
    ├── commons-lang3-3.16.0.jar
    ├── compiler-0.9.14.jar
    ├── jackson-annotations-2.19.0.jar
    ├── jackson-core-2.19.0.jar
    ├── jackson-databind-2.19.0.jar
    ├── jcl-over-slf4j-2.0.17.jar
    ├── jipsy-annotations-1.2.0.jar
    ├── jipsy-processor-1.2.0.jar
    ├── jreleaser-logger-api-1.18.0.jar
    ├── jreleaser-model-api-1.18.0.jar
    ├── jreleaser-resource-bundle-1.18.0.jar
    ├── jreleaser-utils-1.18.0.jar
    ├── slf4j-api-2.0.17.jar
    ├── xz-1.10.jar
    └── zstd-jni-1.5.7-2.jar
    
 which seems wasteful :)
 
 I see two things:
 
  - update jreleaser mechanism to be able to refer to list of jars from the maven repository - jbang can give the list in a json to parse....that avoids copying the same jar around N times and taking up uneccsary space.
  - but even with that; even though above works where my ext use 1.18 and jreleaser is 1.20; i could imagine there being conflicts where `export local` would be better?

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/extensions/Extension.java
maxandersen: i've bumped +1 but also //TODO'ed as i dont have a serial version generator handy

plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/dsl/extensions/Extension.groovy
maxandersen: should i change that in the java code too?

aalmiray: Yes, that should be the case as well.

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/extensions/Extension.java
aalmiray: Version should also be specified otherwise the default bundled version will always be chosen.

I can refactor the code later after merge.

maxandersen: but thats the same for maven ?

maxandersen:
I made https://github.com/jreleaser/jreleaser.github.io/pull/91 as suggestion for docs.

plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/dsl/extensions/Extension.groovy
maxandersen: added commit that change `g|setJBang to `g|setJbang`

PR #1961, opened 2025-08-06
feat: warn on missing variables

### LINKED ISSUE #1960: not helpfull error message when variable does not exist
with https://github.com/jbangdev/jbang-launch/blob/jreleaser-error-msg/jreleaser.yml (that specific tag as i'll probably be trying random things on main) I get this error:

```
JRELEASER_PROJECT_VERSION=1.0.0-SNAPSHOT ./jbang jreleaser@jreleaser full-release --dry-run -scp
🚨 Command flag -scp is deprecated since 1.5.0 and will be removed in 2.0.0. Use --select-current-platform instead
[INFO]  JReleaser 1.19.0
[INFO]  Configuring with jreleaser.yml
[INFO]    - basedir set to /Users/max/code/jbangdev/jbang-launch
[INFO]    - outputdir set to /Users/max/code/jbangdev/jbang-launch/out/jreleaser
[INFO]  Reading configuration
[WARN]  Platform selection is in effect
[WARN]  Artifacts will be filtered by platform matching: [osx-aarch_64]
[INFO]  git-root-search set to false
[INFO]  Loading variables from /Users/max/.jreleaser/config.properties
[INFO]  Validating configuration
[INFO]  Strict mode set to false
Path does not exist: ../../../../../jbang-launch-1.0.0-osx-aarch64.pkg
```

nothing in trace or info on *what* section or path it is trying to resolve

I can kinda deduce it has to do with `distributions` paths which has this:

```
- path: '{{jpackageDir}}/{{distributionName}}-{{projectVersionNumber}}-osx-x86_64.pkg'
        transform: '{{distributionName}}/{{distributionName}}-{{projectEffectiveVersion}}-osx-x86_64.pkg'
        platform: 'osx-x86_64'
```

and I was wondering  where  it got `../../../../../` from ? 

turns out I had copied above with jpackageDir from jreleaser own yaml which "sneakely" has a declaration of jpackageDir on top.

so here is my wish/suggestions:

a) have the error message print out some more context (i.e. "Distribution path `{{jpackageDir}}/{{distributionName}}-{{projectVersionNumber}}-osx-x86_64.pkg` resolved to `../../../../../jbang-launch-1.0.0-osx-aarch64.pkg` does not exist`

b) could there be some warning/error on resolving variables that does not exist? because its   very easy to copy from another jreleaser.yaml with variables that does not exist.

maxandersen:
Fixes #1960  

### Context

prints out warning for missing variables found when evaluating mustache templates.

It is a draft as I quicky realized that to actually log this I would need to have JReleaseLogger available and that would require lots of call point changes so did not want to do that.

so mainly this is just there to show where jreleaser could hook in to help users - whats difficult is that that jreleaser evaluates strings with no context so file etc. info is not very meaningful.

ultimately would be nice to even have ability to fail/error the build if this happens.

### DIFF
--- a/api/jreleaser-model-api/src/main/java/org/jreleaser/mustache/MustacheUtils.java
+++ b/api/jreleaser-model-api/src/main/java/org/jreleaser/mustache/MustacheUtils.java
@@ -17,14 +17,7 @@
  */
 package org.jreleaser.mustache;
 
-import com.github.mustachejava.DefaultMustacheFactory;
-import com.github.mustachejava.Mustache;
-import com.github.mustachejava.MustacheException;
-import com.github.mustachejava.MustacheFactory;
-import com.github.mustachejava.TemplateFunction;
-import org.jreleaser.bundle.RB;
-import org.jreleaser.extensions.api.ExtensionManagerHolder;
-import org.jreleaser.extensions.api.mustache.MustacheExtensionPoint;
+import static org.jreleaser.util.StringUtils.isNotBlank;
 
 import java.io.IOException;
 import java.io.Reader;
@@ -33,10 +26,27 @@ import java.io.StringWriter;
 import java.io.Writer;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.logging.Logger;
 
-import static org.jreleaser.util.StringUtils.isNotBlank;
+import org.jreleaser.bundle.RB;
+import org.jreleaser.extensions.api.ExtensionManagerHolder;
+import org.jreleaser.extensions.api.mustache.MustacheExtensionPoint;
+import org.jreleaser.logging.JReleaserLogger;
+
+import com.github.mustachejava.Binding;
+import com.github.mustachejava.Code;
+import com.github.mustachejava.DefaultMustacheFactory;
+import com.github.mustachejava.Mustache;
+import com.github.mustachejava.MustacheException;
+import com.github.mustachejava.MustacheFactory;
+import com.github.mustachejava.TemplateFunction;
+import com.github.mustachejava.reflect.GuardedBinding;
+import com.github.mustachejava.reflect.MissingWrapper;
+import com.github.mustachejava.reflect.ReflectionObjectHandler;
+import com.github.mustachejava.util.Wrapper;
 
 /**
  * @author Andres Almiray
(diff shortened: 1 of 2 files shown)


maxandersen:
fyi based on https://github.com/spullara/mustache.java/issues/287

api/jreleaser-model-api/src/main/java/org/jreleaser/mustache/MustacheUtils.java
```diff
@@ -138,7 +153,38 @@ private static void applyFunctions(TemplateContext props) {
...
+        return new GuardedBinding(this, name, tc, code) {
+            @Override
+            protected synchronized Wrapper getWrapper(String name, List<Object> scopes) {
+                final Wrapper wrapper = super.getWrapper(name, scopes);
+
+                if (wrapper instanceof MissingWrapper) {
+                    
+                    //TODO: log via JReleaserLogger instead?
+                    //logWarning does not work here because there is no context in the "file" field
+                    //logWarning("Missing variable: " + name, name, scopes, tc);
```
aalmiray, line 172: Does this mean using JReleaserLogger here would result in an exception or missing info?

api/jreleaser-model-api/src/main/java/org/jreleaser/mustache/MustacheUtils.java
```diff
@@ -138,7 +153,38 @@ private static void applyFunctions(TemplateContext props) {
...
+                }
+                return wrapper;
+            }
+        };
+    }
+}
+
     private static class MyMustacheFactory extends DefaultMustacheFactory {
+
+        public MyMustacheFactory(JReleaserLogger logger) {
```
aalmiray, line 183: If the logger arg is not used then this code must be reverted.

api/jreleaser-model-api/src/main/java/org/jreleaser/mustache/MustacheUtils.java
aalmiray: OTOH we could either have a second log file for everything related to mustache templates, or we can keep everything in a single log file in which case JReleaserLogger should be used, even if some context is missing (like the filename because it's not important)

maxandersen: Question for you. I was assuming jreleaser logger was needed to have it end up in trace.log ?

aalmiray: Yes, `trace.log` is the default file output when using the supplied JReleaserLogger

aalmiray:
🎉 This issue has been resolved in `v1.22.0` ([Release Notes](https://github.com/jreleaser/jreleaser/releases/tag/v1.22.0))

PR #1968, opened 2025-08-18
fix: Normalize URL for Nexus3 deployer

ThomasVitale:
Fixes gh-1967

### Context
Add normalization for the Nexus3 URL to ensure the repository trailing slash is included when performing the deploy operation.

### Checklist
- [X] [Review Contribution Guidelines](https://github.com/jreleaser/jreleaser/blob/master/CONTRIBUTING.adoc).
- [X] Make sure all contributed code can be distributed under the terms of the 
      [Apache License 2.0](https://github.com/jreleaser/jreleaser/blob/master/LICENSE), e.g. the code was written by 
      you or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Update [documentation](https://github.com/jreleaser/jreleaser.github.io) when applicable.

Issue #1967, opened 2025-08-18
[nexus3] JReleaser removes trailing slash from URL, leading to 400 error


ThomasVitale:
### Task List

- [X] Steps to reproduce provided
- [X] Stacktrace (if present) provided
- [ ] Example that reproduces the problem (link to git repository is ideal)
- [X] Full description of the issue provided (see below)

### Steps to Reproduce

1. Set up a project with JReleaser using the nexus3 deployer and configure the URL like `https://mynexus.example.com/repositories/my-repo/` (notice the trailing slash).
2. Run `jreleaser config`.
3. In the result, you can see the URL has been changed to `https://mynexus.example.com/repositories/my-repo` (no final trailing slash).
4. Run  `jreleaser full-release`. The task will fail with the following error:

```log
Unexpected error when deploying `com.example.mylibrary:mylibrary-bom`: Error when deploying artifact `com.example.mylibrary:mylibrary-bom`
Got 400 reason: Repository path must have another '/' after initial '/', 
```

Config I'm using:

```
deploy:
  maven:
    nexus2:
      systematic:
        active: SNAPSHOT
        sign: false
        applyMavenCentralRules: true
        url: https://mynexus.example.com/repositories/my-repo/
        snapshotUrl: https://mynexus.example.com/repositories/my-repo/
        snapshotSupported: true
        connectTimeout: 60
        readTimeout: 120
        stagingRepositories:
          - build/staging-deploy
    nexus3:
      systematic:
        active: RELEASE
        sign: false
        applyMavenCentralRules: true
        url: https://mynexus.example.com/repositories/my-repo/
        connectTimeout: 60
        readTimeout: 120
        stagingRepositories:
          - build/staging-deploy
```

I can see the `nexus2` deployer doesn't remove the trailing slash from the `url` property based on what is shown from `jreleaser config`. However, it keeps it for the `snapshotUrl`.

### Expected Behaviour

I would expect the release task to complete successfully.

### Actual Behaviour

The error is the same that happens when trying from a browser to access the repository on the Nexus3 instance. When visiting `https://mynexus.example.com/repositories/my-repo`, it fails with the same error as the one got from JReleaser. When visiting `https://mynexus.example.com/repositories/my-repo/` (notice the trailing slash), it succeeds. 

From `trace.log` (notice the URL is missing the trailing slash):

```log
[...]

PR #1969, opened 2025-08-19
fix: Ensure Nexus3 URLs don't have a trailing slash

ThomasVitale:
### Context

This reverts the change done in https://github.com/jreleaser/jreleaser/pull/1968, which was not correct for Nexus3.

@aalmiray this is awkward. I've been testing both snapshot and production releases across nexus2 and nexus3. It turned out, somehow, I mixed things up, leading to the issue I raised yesterday (https://github.com/jreleaser/jreleaser/issues/1967) which was actually due to misconfiguration. I tested both workflows across both platforms again, and the current version of JReleaser 1.19.0 works correctly.

Nexus3 fails if there's any trailing slash, so it's actually correct that the URL is not normalized. I updated the JavaDoc to explain why it's different from Nexus2.

I'm so sorry about the confusion.

### Checklist


aalmiray:
Thank you 😅

PR #2006, opened 2025-10-31
docs: public keyserver link from http to https

gerardnico:
### Context
The public key download link http://keyserver.ubuntu.com/xxx was not returning any content while the https link does


aalmiray:
Thank you 😄

PR #2043, opened 2026-01-12
feat(announce): Add Twist announcer support

### LINKED ISSUE #1662: [announce] Support Twist as announcer
Developer API docs can be found at https://developer.twist.com/v3/#introduction

shaikhu:
Fixes #1662 

### Context
- Supports creating a thread ([api/v3/threads/add](https://developer.twist.com/v3/#add-thread))
- Supports creating a comment (([api/v3/comments/add](https://developer.twist.com/v3/#add-comment))
- Looking at the [docs](https://developer.twist.com/v3/#testing-purposes), I went with the personal access token for authentication. The interceptor code is cleaner and it matches other announcers e.g.  [LinkedInAnnouncer](https://github.com/jreleaser/jreleaser/blob/22142aa16555890174697108c433fdbd58c8c8ae/sdks/jreleaser-linkedin-java-sdk/src/main/java/org/jreleaser/sdk/linkedin/LinkedinSdk.java#L67). 

Add thread configuration
```
announce:
  twist:
    active: ALWAYS
    accessToken: "<redacted>"
    channelId: 821669
    title: "{{projectName}} {{projectVersion}}"
    message: |
      **{{projectName}} {{projectVersion}} is now available!**
      
      {{projectLongDescription}}
```

Add comment configuration
```
announce:
  twist:
    active: ALWAYS
    accessToken: "<redacted>"
    threadId: 7511087
    message: |
      **{{projectName}} {{projectVersion}} is now available!**
      
      {{projectLongDescription}}
```

Screenshot of example announcement (posting thread + additional comments)
image

### Checklist
- [ :white_check_mark: ] [Review Contribution Guidelines](https://github.com/jreleaser/jreleaser/blob/master/CONTRIBUTING.adoc).
- [ :white_check_mark: ] Make sure all contributed code can be distributed under the terms of the 
- [ :white_check_mark: ] Update [documentation](https://github.com/jreleaser/jreleaser.github.io) when applicable.


core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * @author Usman Shaikh
+ * @since 1.23.0
+ */
+public final class TwistAnnouncer extends AbstractAnnouncer<TwistAnnouncer, org.jreleaser.model.api.announce.TwistAnnouncer> {
+    private static final long serialVersionUID = 2935827447754906932L;
+
+    private String accessToken;
+    private Integer channelId;
```
aalmiray: Should these two ids be treated as strings, but validated as _positive_ numbers?

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
```diff
@@ -0,0 +1,85 @@
...
+            errors.configuration(RB.$("validation_must_not_be_null", "announce.twist.channelId or announce.twist.threadId"));
+        }
+
+        if (announcer.getChannelId() != null && announcer.getThreadId() != null) {
+            errors.configuration(RB.$("validation_exclusive_set", "announce.twist.channelId", "announce.twist.threadId"));
+        }
+
+        if (isNotBlank(announcer.getMessageTemplate()) &&
+            !Files.exists(context.getBasedir().resolve(announcer.getMessageTemplate().trim()))) {
+            errors.configuration(RB.$("validation_directory_not_exist", "twist.messageTemplate", announcer.getMessageTemplate()));
```
aalmiray, line 88: Is this msg bundle key the correct one?

sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/TwistAnnouncer.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return twist.isEnabled();
+    }
+
+    @Override
+    public void announce() throws AnnounceException {
+        if (twist.getChannelId() == null && twist.getThreadId() == null) {
```
aalmiray: This check should be redundant as TwistValidator already checked it.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
```diff
@@ -0,0 +1,85 @@
...
+                listOf(
+                    "announce.twist.access.token",
+                    TWIST_ACCESS_TOKEN),
+                "announce.twist.accessToken",
+                announcer.getAccessToken(),
+                errors,
+                context.isDryrun()));
+
+        if (announcer.getChannelId() == null && announcer.getThreadId() == null) {
+            errors.configuration(RB.$("validation_must_not_be_null", "announce.twist.channelId or announce.twist.threadId"));
```
aalmiray, line 63: Also disable the announcer when a config error is raised, otherwise it may still be enabled when the `yolo` flag is set to true. 

Segue: may need to review other validators to apply the same logic.

sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/TwistAnnouncer.java
```diff
@@ -0,0 +1,114 @@
...
+                context.getLogger().debug(RB.$("twist.message.title"), title);
+                sdk.createThread(twist.getChannelId(), title, message);
+            }
+        } catch (TwistSdkException e) {
+            context.getLogger().trace(e);
+            throw new AnnounceException(e);
+        }
+    }
+
+    private String getResolvedMessage(TemplateContext props) {
```
aalmiray: Why is there a need to resolve the message twice? The announcer (model) already knows how to do it with the `getResolvedMessage` method

aalmiray (changes requested):

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
shaikhu: > Should these two ids be treated as strings, but validated as _positive_ numbers?

It's not clear because I removed it from the pr description, but `accessToken` isn't an integer. It's in the format `oauth2:random_letters_and_numbers`. And `channelId` is an integer, so imo feels wrong storing as a string.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
shaikhu: > Is this msg bundle key the correct one?

I believe so. Same bundle key is used in the [RedditAnnouncerValidator](https://github.com/jreleaser/jreleaser/blob/ca4e1a565981d796fd22b3570e1c82d36eb9695c/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/RedditAnnouncerValidator.java#L106) and [LinkedInAnnouncerValidator](https://github.com/jreleaser/jreleaser/blob/ca4e1a565981d796fd22b3570e1c82d36eb9695c/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/LinkedinAnnouncerValidator.java#L89) when trying to resolve the template.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
aalmiray: Right. The REST API may treat them as ints but the DSL may treat them as strings. My worry is what if the REST API changes type to something that maps beyond Integer.MAX_VALUE. Using string in the DSL accounts for that.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
aalmiray: Will have to check those two. Perhaps the key name is wrong while the value is correct.

sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/TwistAnnouncer.java
shaikhu: > This check should be redundant as TwistValidator already checked it.

Removed.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
shaikhu: > Also disable the announcer when a config error is raised, otherwise it may still be enabled when the `yolo` flag is set to true.

Like [this](https://github.com/jreleaser/jreleaser/compare/103b8afdcecfaeff29ef01d985704220175579d5..36770582f881ff029b50ca90bcfaaa39a8cbc5c5)?

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
shaikhu: > Right. The REST API may treat them as ints but the DSL may treat them as strings. My worry is what if the REST API changes type to something that maps beyond Integer.MAX_VALUE. Using string in the DSL accounts for that.

So IIUC, are you thinking of changing the type of `channelId` and `threadId` to string in the model, and validating they are positive numbers in validator?

sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/TwistAnnouncer.java
shaikhu: > Why is there a need to resolve the message twice? The announcer (model) already knows how to do it with the `getResolvedMessage` method

Removed methods.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
aalmiray: This key is used in many places. The value is correct, thus it stands as suggested.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/announce/TwistAnnouncerValidator.java
aalmiray: Yes, just like that.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
aalmiray: Yes, that's correct.

shaikhu: > Yes, that's correct.

Updated. I left the id's as integers in [ThreadRequest](https://github.com/shaikhu/jreleaser/blob/46c15e0ed6eca693c79d2aa0c0982366768726fe/sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/api/ThreadRequest.java#L28) and [CommentRequest](https://github.com/shaikhu/jreleaser/blob/46c15e0ed6eca693c79d2aa0c0982366768726fe/sdks/jreleaser-twist-java-sdk/src/main/java/org/jreleaser/sdk/twist/api/CommentRequest.java#L28) as these are the JSON payload classes sent to the Twist Rest API.

aalmiray:
@shaikhu I'm looking at this PR and trying to figure out how to yest it on my own. However, I'm stumped on how one can create a Personal Access Token. Do you know how this can be done instead of relying on the test token as linked in the Twist dev documentation?

shaikhu:
> @shaikhu I'm looking at this PR and trying to figure out how to yest it on my own. However, I'm stumped on how one can create a Personal Access Token. Do you know how this can be done instead of relying on the test token as linked in the Twist dev documentation?

Yeah looking through the docs, I don't think it' possible. You have to sign-up, create an application, and then use the associated test token. 

Alternatively, I could try adding you to the 'team'? You then could have access to the channel and token.

aalmiray:
Right. I surmised as much. So, anyone that would want to use this integration would have to do the same, that is, create an app and get a token. 

If so, these steps would have to be added to our documentation.

shaikhu:
> Right. I surmised as much. So, anyone that would want to use this integration would have to do the same, that is, create an app and get a token.
> 
> If so, these steps would have to be added to our documentation.

Tried to highlight this in draft docs pr
https://github.com/jreleaser/jreleaser.github.io/pull/105

Issue #2011, opened 2025-11-06
[deploy] Publishing to Maven Central should be complete at PUBLISHING stage instead of PUBLISHED


aalmiray:
### Discussed in https://github.com/jreleaser/jreleaser/discussions/1879

<div type='discussions-op-text'>

<sup>Originally posted by **wolpert** May 22, 2025</sup>
I've migrated from OSSRH to Maven Central and the times it takes to complete a deployment event is quite large. (30 mins). Before on OSSRH it was just a few mins. After reaching out to Maven Central, this seems to be due to the meaning behind the states provided by the Maven Central API. To quote them:

> Historically, publishing through OSSRH involved a quick upload, close, and then release. At that point, the artifacts would be available from that specific OSSRH server, but they would take ~20 minutes to be fully available on the canonical Maven Central instance. With the Portal, that 20 minutes isn't hidden from you anymore. If your build waits until something reaches the PUBLISHING status, that is the equivalent of a released repository in OSSRH. We are unfamiliar with JReleaser's configuration if they have the ability to poll to that status or not.

My request is to update the plugin to be complete once the status reaches PUBLISHING to cut down on the wait time. I'd love to get feedback from the group on this idea.

For more context, here is a discussion thread I was having: https://github.com/jreleaser/jreleaser/discussions/1876

To make the change, my initial impression is just need to change [this code](https://github.com/jreleaser/jreleaser/blob/4eddcb15e0836b4c924fc4842ff0409d1f028799/sdks/jreleaser-mavencentral-java-sdk/src/main/java/org/jreleaser/sdk/mavencentral/MavenCentral.java#L127):

```
public void publish(String deploymentId) throws MavenCentralException {
    wrap(() -> {
        api.publish(deploymentId);
        waitForState(deploymentId, State.PUBLISHED, State.FAILED);
    });
}
```

Basically, doing the following would fix it (untested):
```
        waitForState(deploymentId, State.PUBLISHED, State.PUBLISHING, State.FAILED);
```

If this makes sense, I can submit a PR</div>

aalmiray:
Proposal: add a flag name `skipPublicationCheck` that is set to `false` by default to keep compatibility with the current behavior. When set to `true` it will allow deployments to continue as soon as any of the following states are reached: `FAILED`, `PUBLISHING`, `PUBLISHED`.

aalmiray:
[...]

PR #2055, opened 2026-01-26
fix(deploy): skipPublicationCheck not stopping at PUBLISHING state

stepan-romankov:
## Summary
Fixes a bug where `skipPublicationCheck = true` does not actually skip waiting for `PUBLISHED` state.

## Problem
When `skipPublicationCheck` is set to `true`, the `waitForState` method correctly includes `PUBLISHING` as an acceptable terminal state:

```java
if (skipPublicationCheck) {
    waitForState(deploymentId, true, State.PUBLISHING, State.PUBLISHED, State.FAILED);
}
```

However, the retry loop uses `Deployment.isTransitioning()` which considers `PUBLISHING` as a transitioning state:

```java
public boolean isTransitioning() {
    return deploymentState == State.PENDING ||
        deploymentState == State.VALIDATING ||
        deploymentState == State.PUBLISHING;  // <-- PUBLISHING is considered transitioning
}
```

This causes the retry loop to continue polling even when `PUBLISHING` state is reached, defeating the purpose of `skipPublicationCheck`.

## Solution
Add an overloaded `isTransitioning(Set<State>)` method that considers acceptable states as terminal:

```java
public boolean isTransitioning(Set<State> acceptableStates) {
    if (acceptableStates.contains(deploymentState)) {
        return false;
    }
    return isTransitioning();
}
```

The `waitForState` method now passes the acceptable states to this new method, allowing the retry loop to stop when `PUBLISHING` is reached if `skipPublicationCheck` is enabled.

## Testing
Added unit tests for the new `isTransitioning(Set<State>)` method covering:
- PUBLISHING state with skipPublicationCheck enabled (should not be transitioning)
- PUBLISHING state with skipPublicationCheck disabled (should be transitioning)
- Other state combinations

## Related
This is a follow-up fix to #2011 which added the `skipPublicationCheck` configuration option.

### DIFF
--- a/sdks/jreleaser-mavencentral-java-sdk/src/main/java/org/jreleaser/sdk/mavencentral/MavenCentral.java
+++ b/sdks/jreleaser-mavencentral-java-sdk/src/main/java/org/jreleaser/sdk/mavencentral/MavenCentral.java
@@ -183,8 +183,10 @@ public class MavenCentral {
     private void waitForState(String deploymentId, boolean checkTimeout, State... states) throws MavenCentralException {
         context.getLogger().debug(RB.$("maven.central.wait.deployment.state", deploymentId, Arrays.asList(states)));
 
+        Set<State> acceptableStates = new LinkedHashSet<>(Arrays.asList(states));
         Boolean[] timeout = new Boolean[]{false};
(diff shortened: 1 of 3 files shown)


aalmiray:
Thank you 😄

PR #2071, opened 2026-02-16
feat(announce): Add Twist announcer support

### LINKED ISSUE #1662: [announce] Support Twist as announcer

shaikhu:
Fixes #1662 

### Context
- Supports posting to a channel using a channel webhook [integration](https://developer.twist.com/v3/#channel-updates).
- Supports replying to a thread using thread webhook [integration](https://developer.twist.com/v3/#thread-updates).

#### JReleaser config file
```
announce:
  twist:
    active: ALWAYS
    webhook: https://twist.com/api/v3/integration_incoming/post_data?install_id=redacted&install_token=redacted
    messageTemplate: src/jreleaser/templates/twist-channel.tpl
```

#### twist-channel.tpl 
```
{
  "content": "**{{projectName}} {{projectVersion}} is now available!**\n\n{{projectLongDescription}}\n\nReleased with JReleaser",
  "title": "{{projectName}} {{projectVersion}} is now available!"
}
```

#### twist-thread.tpl
```
{
  "content": "**{{projectName}} {{projectVersion}} is now available!**\n\n{{projectLongDescription}}\n\nReleased with JReleaser",
}
```

### Checklist
- [✔️] [Review Contribution Guidelines](https://github.com/jreleaser/jreleaser/blob/master/CONTRIBUTING.adoc).
- [✔️] Make sure all contributed code can be distributed under the terms of the 

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
@@ -1,0 +1,190 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Copyright 2020-2026 The JReleaser authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jreleaser.model.internal.announce;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.jreleaser.bundle.RB;
+import org.jreleaser.model.Active;
+import org.jreleaser.model.Constants;
+import org.jreleaser.model.JReleaserException;
+import org.jreleaser.model.internal.JReleaserContext;
+import org.jreleaser.mustache.TemplateContext;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Path;
+import java.util.Map;
+
+import static java.util.Collections.unmodifiableMap;
+import static org.jreleaser.model.Constants.HIDE;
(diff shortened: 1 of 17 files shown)


core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TwistAnnouncer.java
```diff
@@ -0,0 +1,190 @@
...
+        WebhookAnnouncer announcer = new WebhookAnnouncer();
+        announcer.setName(getName());
+        announcer.setWebhook(webhook);
+        announcer.setMessageTemplate(messageTemplate);
+        announcer.setStructuredMessage(false);
+        announcer.setConnectTimeout(getConnectTimeout());
+        announcer.setReadTimeout(getReadTimeout());
+        announcer.setExtraProperties(getExtraProperties());
+        return announcer;
+    }
```
shaikhu, line 189: I originally attempted the same approach used by [DiscordAnnouncer](https://github.com/jreleaser/jreleaser/blob/main/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/DiscordAnnouncer.java#L218..L219) e.g. `setStructuredMessage(true)` and `setMessageProperty("content")`. While this worked for the thread integration, it wasn't enough for the channel integration, which has the additional property `title`. To resolve this I used the same approach used by the [TeamsAnnouncer](https://github.com/jreleaser/jreleaser/blob/main/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/announce/TeamsAnnouncer.java#L188..L198) e.g. user has to enter the content (valid json) through the `messageTemplate` file.

shaikhu:
Thanks for the review (and patience). While getting ready to write the docs I've reliased the `Discord`, `GoogleChat`, and `Teams` Announcers are all deprecated. The recommended approach is to use the WebHooks announcer directly. See docs [here](https://jreleaser.org/guide/latest/reference/announce/webhooks.html#_teams).

Using the below snippet while setting `JRELEASER_TWIST_WEBHOOK` environment variable works with the existing code on `main`.
```
announce:
  webhooks:
    twist:
      active: ALWAYS
      messageTemplate: src/jreleaser/templates/twist-channel.tpl
```

I don't think it makes sense to write a new `TwistAnnouncer` while the existing code can be used. 

If you would like the support for Twist to be documented I could raise a PR to add Twist to the WebHooks [examples](https://jreleaser.org/guide/latest/reference/announce/webhooks.html#_examples) ?

aalmiray:
Right. If there's no need to add code for the announcer and rather define the convention required with the current webhooks announcer, then we only need an update in the docs

PR #2073, opened 2026-02-18
feat(changelog): support path filtering for monorepo changelog generation

### LINKED ISSUE #2072: [release] Support path filtering for changelog generation in monorepos
When using JReleaser in a monorepo with multiple projects, the changelog includes commits from the entire repository rather than just the project being released.

JGit's `LogCommand` supports `.addPath()` for filtering commits by path. It would be useful to expose this in the changelog configuration:

```yaml
release:
  github:
    changelog:
      paths:
        - 'my-subproject/'
```

Currently I'm working around this with an external changelog generated via `git log ... -- .`, but native support would be cleaner.

jruaux:
When using JReleaser in a monorepo with multiple projects, the changelog includes commits from the entire repository rather than just the project being released.

This PR adds a `paths` property to the changelog configuration that filters commits to only those affecting the specified paths using JGit's `LogCommand.addPath()`.

**Usage:**
```yaml
release:
  github:
    changelog:
      paths:
        - 'my-subproject/'
```

Closes #2072

### DIFF
--- a/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/release/Changelog.java
+++ b/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/release/Changelog.java
@@ -63,6 +63,8 @@ public interface Changelog extends Domain, EnabledAware, ExtraProperties {
 
     String getPreset();
 
+    Set<String> getPaths();
+
     Hide getHide();
 
     Contributors getContributors();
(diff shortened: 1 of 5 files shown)


core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/release/Changelog.java
```diff
@@ -69,6 +69,7 @@ public final class Changelog extends AbstractModelObject<Changelog> implements D
     private final Hide hide = new Hide();
     private final Contributors contributors = new Contributors();
     private final Append append = new Append();
+    private final Set<String> paths = new LinkedHashSet<>();
```
aalmiray, line 72: The serialVersionUID must be updated as well

api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/release/Changelog.java
```diff
@@ -63,6 +63,8 @@ public interface Changelog extends Domain, EnabledAware, ExtraProperties {
 
     String getPreset();
 
+    Set<String> getPaths();
```
aalmiray, line 66: The DSL in the plugins/jreleaser-gradle-plugin should be updated as well

aalmiray:
Thank you 😁

PR #2076, opened 2026-02-26
Fix misleading message from disabled deployer

gastaldi:
The reason (`because project is snapshot`) is not true when the disabled deployer is configured as `active: SNAPSHOT` and the project is NOT a SNAPSHOT
eg.
```
[WARN]    [validation] Deployer nexus2.maven-central disabled because project is snapshot
```


aalmiray (approved):

aalmiray:
Thank you 😄

PR #2080, opened 2026-03-06
fix: use multi-stage build in Docker templates to avoid duplicate layers

### LINKED ISSUE #2079: [package] Docker template for JAVA_BINARY creates duplicate layers
## Problem

The default `Dockerfile.tpl` for `JAVA_BINARY` distributions creates duplicate layers:

```dockerfile
COPY assembly/ /                    # Layer 1: contains ZIP

RUN unzip ... && rm ...             # Layer 2: contains extracted files
```

Docker layers are immutable. The `rm` command does not reduce image size because the ZIP is already in Layer 1.

For a 750MB ZIP, this results in a ~1.5GB image instead of ~750MB.

## Suggested Fix

Use a multi-stage build:

```dockerfile
FROM alpine AS extractor
RUN apk add --no-cache unzip
COPY assembly/ /tmp/
RUN unzip /tmp/{{distributionArtifactFile}} -d /opt

FROM {{dockerBaseImage}}
COPY --from=extractor /opt/{{distributionArtifactRootEntryName}} /{{distributionArtifactRootEntryName}}
```

This ensures the ZIP file is not present in the final image.

## Affected Template

`Dockerfile.tpl` for `JAVA_BINARY` distributions (local artifact mode).

The remote template (`Dockerfile-remote.tpl`) does not have this issue because it downloads and extracts in a single `RUN` command.

jruaux:
## Problem

Docker templates create duplicate layers because they `COPY` files then `RUN` commands that move/delete them. Docker layers are immutable, so the original files remain in the first layer.

For a 750MB artifact, this doubles the image size.

### java-binary and jlink

Use multi-stage build: extract in throwaway Alpine stage, copy only extracted files to final image.

### single-jar

Simplified approach: copy JAR directly to `/app/`, use `java -jar` entrypoint.

The bin/lib directory structure with launcher script is unnecessary for Docker:
- Launcher script is useful for bare-metal (finding JAVA_HOME, env setup)
- In containers: Java location is known, ENTRYPOINT defines how to run
- No file reorganization = no duplicate layers = no multi-stage needed

## Changes

| Template | Approach |
|----------|----------|
| `java-binary` | Multi-stage build with Alpine extractor |
| `jlink` | Multi-stage build with Alpine extractor |
| `single-jar` | Direct COPY, no bin/lib structure |

Fixes #2079

### DIFF
--- a/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/binary/docker/Dockerfile.tpl
+++ b/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/binary/docker/Dockerfile.tpl
@@ -1,4 +1,12 @@
 # {{jreleaserCreationStamp}}
+# Multi-stage build to avoid duplicate layers
+
+FROM alpine:latest AS assembler
+
+COPY assembly/ /assembly/
+
+RUN chmod +x /assembly/{{distributionArtifactRootEntryName}}/bin/{{distributionExecutableUnix}}
+
 FROM {{dockerBaseImage}}
 
 {{#dockerLabels}}
@@ -9,8 +17,7 @@ LABEL {{.}}
 {{.}}
 {{/dockerPreCommands}}
 
-COPY assembly/ /
-RUN chmod +x {{distributionArtifactRootEntryName}}/bin/{{distributionExecutableUnix}}
+COPY --from=assembler /assembly/ /
 
 {{#dockerPostCommands}}
 {{.}}
--- a/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/flat-binary/docker/Dockerfile.tpl
+++ b/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/flat-binary/docker/Dockerfile.tpl
@@ -1,4 +1,12 @@
 # {{jreleaserCreationStamp}}
+# Multi-stage build to avoid duplicate layers
+
+FROM alpine:latest AS assembler
+
+COPY assembly/ /assembly/
+
+RUN chmod +x /assembly/{{distributionExecutableName}}
+
 FROM {{dockerBaseImage}}
 
 {{#dockerLabels}}
@@ -9,8 +17,7 @@ LABEL {{.}}
 {{.}}
 {{/dockerPreCommands}}
 
-COPY assembly/ /
-RUN chmod +x /{{distributionExecutableName}}
+COPY --from=assembler /assembly/ /
 
 {{#dockerPostCommands}}
 {{.}}
--- a/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/native-image/docker/Dockerfile.tpl
+++ b/core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/native-image/docker/Dockerfile.tpl
@@ -1,4 +1,12 @@
 # {{jreleaserCreationStamp}}
+# Multi-stage build to avoid duplicate layers
+
+FROM alpine:latest AS assembler
+
+COPY assembly/ /assembly/
+
+RUN chmod +x /assembly/{{distributionArtifactRootEntryName}}/bin/{{distributionExecutableUnix}}
+
 FROM {{dockerBaseImage}}
 
 {{#dockerLabels}}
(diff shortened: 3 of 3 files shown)


aalmiray:
Thank you for the PR. I think the BINARY templates would need a similar update. 

Also, there may be additional files that must be copied to the container image such as a readme, license, etc, not just the matching distribution archive. Thus, the use of `COPY /assembly /`.

core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/java-binary/docker/Dockerfile.tpl
```diff
@@ -1,4 +1,16 @@
 # {{jreleaserCreationStamp}}
+# Multi-stage build to avoid duplicate layers
+
+FROM alpine:latest AS extractor
```
aalmiray, line 4: Is it a problem that the build image is based on Alpine but the base image may not (could be Ubuntu for example)

core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/single-jar/docker/Dockerfile.tpl
```diff
@@ -11,19 +11,11 @@ LABEL {{.}}
...
-    mv /{{distributionArtifactFile}} /{{distributionName}}-{{projectVersion}}/lib
-
-ENV PATH="${PATH}:/{{distributionName}}-{{projectVersion}}/bin"
-
 {{#dockerPostCommands}}
 {{.}}
 {{/dockerPostCommands}}
 
-ENTRYPOINT ["/{{distributionName}}-{{projectVersion}}/bin/{{distributionExecutableUnix}}"]
+ENTRYPOINT {{dockerEntrypoint}}
```
aalmiray: Is the value of this variable set to the correct thing now? I suppose the docker packager validator should make sure it has the correct value.

core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/java-binary/docker/Dockerfile.tpl
jruaux: No, this is not a problem. The extractor stage only runs `unzip` which produces platform-independent output (the extracted files are Java bytecode, shell scripts, etc.). The `COPY --from=extractor` instruction just copies these files to the final image - it doesn't copy any Alpine-specific binaries. The extractor image is discarded after the build.

core/jreleaser-templates/src/main/resources/META-INF/jreleaser/templates/single-jar/docker/Dockerfile.tpl
jruaux: Good catch! I've now applied the same multi-stage build approach to single-jar. The original template had the same duplicate layer issue - `mv` doesn't remove files from previous Docker layers, so the JAR was duplicated. Now it uses a multi-stage build that maintains the expected bin/lib directory structure and hardcoded ENTRYPOINT.

aalmiray: Great, thanks!

I suppose every Dockerfile template hast to be reviewed. Please take one more view at every single one in case there may be any adjustments needed, thank you.

jruaux: Done! I've reviewed and updated all 6 Docker templates to use multi-stage builds:

| Template | Issue |
|----------|-------|
| **java-binary** | `unzip` + `rm` duplicated extracted content |
| **jlink** | `unzip` + `rm` duplicated extracted content |
| **single-jar** | `mv` duplicated JAR file |
| **binary** | `chmod` duplicated binary |
| **flat-binary** | `chmod` duplicated binary |
| **native-image** | `chmod` duplicated binary |

Testing confirmed that `chmod` also duplicates files in Docker layers - a 50MB binary with `chmod` in a RUN layer resulted in 118MB image size vs 65.8MB with multi-stage (45% reduction).

aalmiray:
Thank you 😄

PR #2087, opened 2026-03-15
feat(maven): add download timeout support to jdks-maven-plugin

### LINKED ISSUE #2083: [jdks] Fix timeout for jdk download
When I try to download a jdk using setup-disco, I get a `Read timed out` error.

I have tried the `readTimeout` property found in the JReleaser documentation but it seems that `SetupDiscoMojo.java` is not using it when downloading a jdk:

https://github.com/jreleaser/jreleaser/blob/77d1e3c2492256a4da9d23161162599b4bc9f08a/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/JdkHelper.java#L111-L126

PrakarshSrivastav:
Fixes #2083

### Checklist

### DIFF
--- a/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/AbstractJdksMojo.java
+++ b/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/AbstractJdksMojo.java
@@ -34,6 +34,12 @@ abstract class AbstractJdksMojo extends AbstractSetupMojo {
     @Parameter(required = true)
     protected List<Jdk> jdks;
 
+    @Parameter(property = "jdks.setup.connect.timeout")
+    protected int connectTimeout;
+
+    @Parameter(property = "jdks.setup.read.timeout")
+    protected int readTimeout;
+
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
         Banner.display(project, getLog());
@@ -47,6 +53,13 @@ abstract class AbstractJdksMojo extends AbstractSetupMojo {
     protected abstract void doExecute() throws MojoExecutionException;
 
     protected void validate() throws MojoFailureException {
+        if (connectTimeout <= 0 || connectTimeout > 300) {
+            connectTimeout = 20;
+        }
+        if (readTimeout <= 0 || readTimeout > 300) {
+            readTimeout = 60;
+        }
+
         if (null == jdks || jdks.isEmpty()) return;
 
         Errors errors = new Errors();
--- a/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/JdkHelper.java
+++ b/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/JdkHelper.java
@@ -58,16 +58,21 @@ public class JdkHelper {
     private final MavenSession session;
     private final BuildPluginManager pluginManager;
     private final ArchiverManager archiverManager;
+    private final int connectTimeout;
+    private final int readTimeout;
 
     public JdkHelper(MavenProject project, Log log, File outputDirectory,
                      MavenSession session, BuildPluginManager pluginManager,
-                     ArchiverManager archiverManager) {
+                     ArchiverManager archiverManager,
+                     int connectTimeout, int readTimeout) {
         this.project = project;
         this.log = log;
         this.outputDirectory = outputDirectory;
         this.session = session;
         this.pluginManager = pluginManager;
         this.archiverManager = archiverManager;
+        this.connectTimeout = connectTimeout;
+        this.readTimeout = readTimeout;
     }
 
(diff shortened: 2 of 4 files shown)


aalmiray (approved):

PrakarshSrivastav:
I think you will have to merge this PR as I don't have commit access. Thanks!

plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/AbstractJdksMojo.java
```diff
@@ -47,6 +53,13 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     protected abstract void doExecute() throws MojoExecutionException;
 
     protected void validate() throws MojoFailureException {
+        if (connectTimeout <= 0 || connectTimeout > 300) {
+            connectTimeout = 20;
+        }
+        if (readTimeout <= 0 || readTimeout > 300) {
+            readTimeout = 60;
+        }
+
```
mhoffrog, line 62: @PrakarshSrivastav This implementation leads to an irritating user experience in case of configuring beyond upper limit of this validation. Limiting on 300s timeout and if exceeding then replacing by 20 or 60 seconds - without any warning. I'd recommend to match the value for replacing with the value limiting to AND log a warning in case of limiting the configuration with logging the value effectively being replaced.

PrakarshSrivastav: Thanks for pointing this out. That’s a good point about the UX and the silent replacement of the configured values. I’m not able to update the implementation right now, but I’ll take another look at it and adjust the logic to cap the values and add a warning log in the morning.

aalmiray: @mhoffrog the suggested vale replacements occur somewhere else in the codebase where timeouts are validated, and thus these settings are consistent. Also, timeouts are set to minutes, not seconds. Actual values are multiplied by 1000 at the time of usage.

mhoffrog: > ... timeouts are set to minutes, not seconds. Actual values are multiplied by 1000 at the time of usage.

@aalmiray Hmm - are you really sure about the minutes? The values at the codeplace I commented are in seconds - multiplying by 1000 is done, since the used API expects the value in milliseconds.

aalmiray:
Thank you 😄

plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/AbstractJdksMojo.java
aalmiray: Hmm right. So it resolves to 20s for connection and 5 mins for read

mhoffrog: This is not as it is currently coded.
Current implementation will enforce the value configured as follows:
| Configured:  | <= 0 | from 1s to 300s | > 300s |
| ---:                | ---     | ---                      | --- |
| `connectTimeout` | 20s | as configured | 20s |
| `readTimeout` | 60s | as configured | 60s |

To me it is strange that in case the configuration exceeds 300s (=5 minutes) the effective value gets enforced to a somehow arbitrary defined value.
In case of cutting to 5 minutes max, I would expect that the effective value enforced would be 300s. In case getting the connectTimeout below 20s or the readTimeout below 60s - I would expect that these cases would be the low cut value. But instead you can configure below those values unless you do not configure below or equal to 0.

aalmiray: Oh gees, I see it now. I'm sorry, I thought this logic seemed so similar to the other places that I didn't pay much attention to it. These values need to be changed indeed. 20 & 60 should be the minimum, while 300 the maximum.

mhoffrog: OK, we are on same page now!
In addition I'd recommend to log an appropriate  warning, in case of any overriding/cutting of an original configuration to make the user aware of an unintended configuration.

aalmiray: In that case there would have to be several places where such log statements must be added. In any case, actual values are shown in the fully resolved model found in the trace log.

aalmiray:
🎉 This issue has been resolved in `v1.24.0` ([Release Notes](https://github.com/jreleaser/jreleaser/releases/tag/v1.24.0))

PR #2078, opened 2026-03-04
feat(gradle): Fix access to Task.project at task execution time

mhoffrog:
Fixes #1992

### Context
The Gradle Project instance object must not be accessed from task execution during Gradle build runtime phase.
This is violating Gradle configuration cache pre-requisites and it is currently raising a warning since this kind of access is deprecated and will run into an error with Gradle 10.

### Solution
* **Refactor**
  * Internal plugin architecture reworked to centralize Gradle project metadata capture and configuration, reducing direct Gradle API coupling.
  * Tasks now share a common base for consistent initialization, simplifying task configuration and logging.
  * Several task inputs/outputs were streamlined; environment task now always re-runs (up-to-date disabled) and environment display uses the task's project directory.
  * Adds a JReleaserGradleProjectCapture value object and refactors the plugin to use it instead of direct Gradle Project access; method signatures and task base classes were updated to accept/use the capture and centralize project property resolution.

### Recommendations to do
- Class `AbstractJReleaserTask` should be renamed to something like `AbstractJReleaserCommonOptionsTask` or `AbstractJReleaserCoreOptionsTask` since it is now inheriting from `AbstractJReleaserDefaultTask` as next in the tasks inheritance hierarchy.

### Checklist
- [x] [Review Contribution Guidelines](https://github.com/jreleaser/jreleaser/blob/master/CONTRIBUTING.adoc).
- [x] Make sure all contributed code can be distributed under the terms of the 
      [Apache License 2.0](https://github.com/jreleaser/jreleaser/blob/master/LICENSE), e.g. the code was written by 
      you or the original code is licensed under [a license compatible to Apache License 2.0](https://apache.org/legal/resolved.html).
- [ ] Update [documentation](https://github.com/jreleaser/jreleaser.github.io) when applicable.

PR #2093, opened 2026-03-18
feat(gradle): Rename AbstractJReleaserTask to AbstractJReleaserCommonOptionsTask

mhoffrog:
Following up an open to do recommendation of #2078

### Context

This is a cosmetic change to reflect better what the former `AbstractJReleaserTask` class is being supposed for.
👉 This change is preserving the git history of the original source file.

### Checklist

### DIFF
--- a/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/tasks/AbstractJReleaserCommonOptionsTask.groovy
+++ b/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/tasks/AbstractJReleaserCommonOptionsTask.groovy
@@ -18,19 +18,14 @@
 package org.jreleaser.gradle.plugin.tasks
 
 import groovy.transform.CompileStatic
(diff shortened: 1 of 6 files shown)


aalmiray:
Hmm don't know if this change is actually needed at this point. AbstractJReleaserTask continues to be the base type, the fact that it now adds more options/behavior to its subclasses shouldn't warrant a name change, should it?

mhoffrog:
It is up to you - what you prefer. My thought was that since the newly introduced `AbstractJReleaserDefaultTask` did become the overall tasks root base class now and `AbstractJReleaserTask` is now inheriting from it. So the name should reflect that it is not the neutral root base class anymore. The name with the options came to my mind, since it comes with a couple of options that are shared with many of the tasks classes inheriting from it - and this was not changed versus before content.

aalmiray:
Understood. It would make sense to rename if the newly added `AbstractJReleaserDeaultTask` was renamed to `AbstractJReleaserTask` as well. I think it's not worth performing this action at the moment.

Thank you for bringing it up though.

PR #2121, opened 2026-05-19
fix(assemble): drop literal quotes around jlink --add-options value

### LINKED ISSUE #2120: [assemble] jlink --add-options value wrapped in literal quotes breaks the runtime image
### Task List

- [x] Stacktrace (if present) provided

### Steps to Reproduce

1. Configure a jlink assembler with two or more `jvmOptions.universal` entries, e.g.:

   ```yaml
   assemble:
     jlink:
       myapp:
         active: ALWAYS
         java:
           jvmOptions:
             universal:
               - '-XX:+IgnoreUnrecognizedVMOptions'
               - '--enable-native-access=ALL-UNNAMED'
   ```

2. Assemble: `jreleaser assemble --assembler jlink`.
3. From the produced runtime image, run any command that starts the JVM — for example `bin/java --version`.

### Expected Behaviour

The JVM starts and applies each option separately, as if they had been passed individually on the command line.

### Actual Behaviour

```
$ ./bin/java --version
Unrecognized VM option 'IgnoreUnrecognizedVMOptions --enable-native-access=ALL-UNNAMED'
Did you mean '(+/-)IgnoreUnrecognizedVMOptions'?
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

Every invocation of the bundled `bin/java` (and therefore the generated `bin/<distribution>` launcher) fails — even the trivial `--version` call shown above — because the JVM treats all four options as one malformed `-XX:+...` argument.

### Root cause

`core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JlinkAssemblerProcessor.java:289` builds the jlink argument like this:

```java
cmd.arg("--add-options=\"" + join(" ", assembler.getJava().getJvmOptions().getResolvedUniversal(context)) + "\"");
```

The `\"...\"` characters are literal. JReleaser invokes jlink via `ProcessBuilder` (no shell), so the embedded double quotes become part of the option value that jlink writes into the runtime image. At JVM startup the launcher reads back the option value with the quotes included and treats the whole quoted string as a single malformed `-XX:+...` option, producing the error above.

### Minimal reproducer against raw jlink (no JReleaser)

```sh
JDK=/path/to/openjdk-25

# ✅ Works — no embedded literal quotes
$JDK/bin/jlink --module-path "$JDK/jmods" --add-modules java.base \
  --add-options='-XX:+IgnoreUnrecognizedVMOptions --enable-native-access=ALL-UNNAMED' \
  --output good
./good/bin/java --version

# ❌ Fails — value wrapped in literal double quotes (what JReleaser produces)
$JDK/bin/jlink --module-path "$JDK/jmods" --add-modules java.base \
  '--add-options="-XX:+IgnoreUnrecognizedVMOptions --enable-native-access=ALL-UNNAMED"' \
  --output bad
./bad/bin/java --version
# Unrecognized VM option 'IgnoreUnrecognizedVMOptions --enable-native-access=ALL-UNNAMED'
```

### Suggested fix

Drop the literal `\"...\"` characters at `JlinkAssemblerProcessor.java:289` so jlink receives the unquoted, space-separated option list it expects. PR to follow.

### Environment Information

- **Operating System:** macOS 25.4 (arm64); also reproduced on Linux x86_64 (Amazon Corretto 17 + Zulu 25 jlink-bundled runtime)
- **JReleaser Version:** 1.23.0 and 1.24.0 (line is unchanged on `main` as of 03ece6e)
- **JDK Version:** OpenJDK 25.0.2 and Azul Zulu 25.0.1 (the bug reproduces with any jlink from JDK 17+)

jruaux:
## Summary

Fixes #2120.

When a jlink assembler defines `jvmOptions.universal`, JReleaser builds the jlink command with literally-quoted value:

```java
cmd.arg("--add-options=\"" + join(" ", ...getResolvedUniversal(context)) + "\"");
```

The arg is passed to jlink via `ProcessBuilder` (no shell), so the `"` characters become part of the option value stored in the runtime image. At JVM startup the launcher reads the value back with the quotes included and treats the entire joined option string as one malformed `-XX:+...` argument. The result is that **every** invocation of the bundled `bin/java` fails — including the trivial `bin/java --version`.

This PR removes the literal quotes so jlink receives the unquoted, space-separated option list it expects.

```diff
-            cmd.arg("--add-options=\"" + join(" ", assembler.getJava().getJvmOptions().getResolvedUniversal(context)) + "\"");
+            cmd.arg("--add-options=" + join(" ", assembler.getJava().getJvmOptions().getResolvedUniversal(context)));
```

The minimal reproducer (no JReleaser involved) and full root-cause analysis are in #2120.

## Test plan

- [x] Reproduced the failure with raw `jlink` by passing `'--add-options="...multiple options..."'` (literal quotes) — `bin/java --version` fails with `Unrecognized VM option '...'`.
- [x] Verified the same `jlink` invocation without the literal quotes produces a runtime where `bin/java --version` and the generated launcher work correctly.
- [x] Verified end-to-end against a real downstream project (riotx) with multiple universal `jvmOptions` (`-XX:+IgnoreUnrecognizedVMOptions`, `--add-opens=java.base/java.nio=ALL-UNNAMED`, `--enable-native-access=ALL-UNNAMED`, `--sun-misc-unsafe-memory-access=allow`): after applying this patch, `bin/java --version` and `bin/<distribution> --version` both succeed.

### DIFF
--- a/core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JlinkAssemblerProcessor.java
+++ b/core/jreleaser-engine/src/main/java/org/jreleaser/assemblers/JlinkAssemblerProcessor.java
@@ -286,7 +286,7 @@ public class JlinkAssemblerProcessor extends AbstractAssemblerProcessor<org.jrel
             .arg("--add-modules")
             .arg(join(",", moduleNames));
         if (assembler.getJava().getJvmOptions().isSet()) {
-            cmd.arg("--add-options=\"" + join(" ", assembler.getJava().getJvmOptions().getResolvedUniversal(context)) + "\"");
+            cmd.arg("--add-options=" + join(" ", assembler.getJava().getJvmOptions().getResolvedUniversal(context)));
         }
         cmd.arg("--output")
             .arg(maybeQuote(imageDirectory.toString()));


aalmiray (approved):

aalmiray:
Thank you 😄

PR #2127, opened 2026-05-22
fix(jdks): drop unsupported connectTimeOut element

### LINKED ISSUE #2126: [jdks] Fix connectTimeout for jdk download
`setup-disco` goal is broken in v1.24.0

It seems that the underlying plugin (`io.github.download-maven-plugin:download-maven-plugin:2.1.0`) doesn't support `connectTimeout` parameter.

```
[ERROR] Failed to execute goal org.jreleaser:jdks-maven-plugin:1.24.0:setup-disco (download-jre) on project jdemetra-app-snapshot: Unable to execute mojo: Unable to parse configuration of mojo io.github.download-maven-plugin:download-maven-plugin:2.1.0:wget for parameter connectTimeOut: Cannot find 'connectTimeOut' in class io.github.download.maven.plugin.internal.WGetMojo -> [Help 1]
```

Bug introduced in https://github.com/jreleaser/jreleaser/pull/2087/changes#diff-035349c56da14714b170ca52d06d3123014053672f734369120f70dbb264e296 for the issue https://github.com/jreleaser/jreleaser/issues/2083

Related code in `download-maven-plugin`: https://github.com/download-maven-plugin/download-maven-plugin/blob/e8c1b68c4bd5338fab1a77b9ca56430de8b498c6/src/main/java/io/github/download/maven/plugin/internal/WGetMojo.java#L807

mvanhorn:
Fixes #2126

## Summary

The `setup-disco` goal is currently broken in v1.24.0 because `JdkHelper` passes a `connectTimeOut` element to `download-maven-plugin` 2.1.0, but `WGetMojo` does not declare that parameter. Maven's configuration parser rejects the build with `Cannot find 'connectTimeOut' in class ... WGetMojo`. Drop the element and pass `Math.max(connectTimeout, readTimeout) * 1000` as `readTimeOut` so neither caller-set value is silently dropped.

### Context

`download-maven-plugin` 2.1.0 [`WGetMojo.java:807-808`](https://github.com/download-maven-plugin/download-maven-plugin/blob/e8c1b68c4bd5338fab1a77b9ca56430de8b498c6/src/main/java/io/github/download/maven/plugin/internal/WGetMojo.java#L807) wires `readTimeOut` into BOTH the HTTP connect timeout and the socket timeout:

```java
.withConnectTimeout(this.readTimeOut)
.withSocketTimeout(this.readTimeOut)
```

There is no separate `connectTimeOut` parameter in that plugin version, so PR #2087 introduced a configuration shape the plugin cannot parse. Every `setup-disco` invocation against v1.24.0 hard-fails at goal execution time.

## Why this matters

Two valid fixes were available:

1. Just drop the bad element. The user-facing JReleaser `connectTimeout` knob still validates and clamps via `AbstractJdksMojo`, but it gets silently ignored for the download step (because the plugin reuses `readTimeOut` internally for connect anyway).
2. Drop the bad element AND pass `Math.max(connectTimeout, readTimeout) * 1000` as `readTimeOut`. The larger user-set value becomes the effective combined timeout. Neither knob is silently dropped.

This PR takes option 2 because the JReleaser docs/parameter validation (`MINIMUM_CONNECT_TIMEOUT = 5`, `MAXIMUM_CONNECT_TIMEOUT = 300` in `AbstractJdksMojo`) imply the value is supposed to do something. Without that change, a user who set `connectTimeout = 60` and left `readTimeout` at its default (smaller) value would see their timeout silently ignored.

## Changes

- `JdkHelper.java:127` - remove the `connectTimeOut` element entirely (download-maven-plugin doesn't accept it).
- `JdkHelper.java:128` - change `readTimeout` to `Math.max(connectTimeout, readTimeout)` so the larger user-set knob wins, with a comment explaining the rationale.
- No public API change. `AbstractJdksMojo.connectTimeout` / `readTimeout` parameters and their validation are unchanged.

## How to test

- `./gradlew :jdks-maven-plugin:compileJava` passes locally.
- The original `Cannot find 'connectTimeOut'` error is rooted in maven-plugin configuration parsing, not Java compile; the fix verifiably removes the unparseable element name so the parse error class is structurally eliminated.

### Checklist

- [x] All contributed code is distributable under Apache License 2.0 (this is a small edit to an existing Apache-2.0 file).

AI-assisted.

### DIFF
--- a/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/JdkHelper.java
+++ b/plugins/jdks-maven-plugin/src/main/java/org/jreleaser/jdks/maven/plugin/JdkHelper.java
@@ -124,8 +124,11 @@ public class JdkHelper {
                     element("outputDirectory", jdkExtractDirectory.getAbsolutePath()),
                     element("cacheDirectory", cacheDirectory),
                     element("outputFileName", filename),
-                    element("connectTimeOut", String.valueOf(connectTimeout * 1000)),
-                    element("readTimeOut", String.valueOf(readTimeout * 1000))
+                    // download-maven-plugin 2.1.0 does not expose a separate connect timeout
+                    // parameter; WGetMojo wires readTimeOut into both connect and socket
+                    // timeouts internally. Pass the larger of the two so neither caller-set
+                    // value is silently dropped.
+                    element("readTimeOut", String.valueOf(Math.max(connectTimeout, readTimeout) * 1000))
                 ),
                 executionEnvironment(
                     project,


aalmiray (approved):

aalmiray:
@mvanhorn please review and sign the CLA if you agree

aalmiray:
It seems this PR mixed a previous commit related to the JLinkAssembler. Could you recheck and resubmit? Thank you.

mvanhorn:
Cleaned up in 607e547d3. This PR now contains only the connectTimeOut fix in JdkHelper.java. The `jlink --add-options` commit that had snuck onto the branch was @jruaux's #2121, which is already merged, so I dropped it here rather than resubmit it. Ready for another look.

aalmiray (approved):

aalmiray:
Thank you :smile:

mvanhorn:
Thanks for the quick merge, @aalmiray. Glad the setup-disco fix is in so the connectTimeOut config error stops breaking v1.24.0 builds.

mvanhorn:
Appreciate the merge @aalmiray. Dropping the unsupported connectTimeOut element keeps the JDK config from tripping people up.

mvanhorn:
Appreciate the quick merge @aalmiray, the unsupported connectTimeOut element is gone from the JDK config.

aalmiray:
@mvanhorn you've posted 3 similar comments after the PR was merged. What is going on?

mvanhorn:
Thanks @aalmiray for the quick turnaround on the connectTimeOut cleanup.

PR #2129, opened 2026-05-23
fix(signing): skip signing validation errors when --yolo is set

### LINKED ISSUE #2128: [signing] Signing fails when secret key is missing even when yolo is set
### Task List

- [] Stacktrace (if present) provided

### Steps to Reproduce

1. checkout https://github.com/jreleaser/helloworld-java-bin
2. configure a signing block like this
```
signing:
  pgp:
    active: ALWAYS
```
3. Make sure you don't have JRELEASER_GPG_SECRET_KEY defined
4. `mvnw verify`
5. `jreleaser assemble`
6. `jreleaser release --dry-run --yolo`
### Expected Behaviour

A local release should be performed with signing skipped.

### Actual Behaviour

An exception occurs because the PGP secret key is missing

### Environment Information

- **Operating System**: N/A
- **JReleaser Version:** 1.24.0
- **JDK Version:** N/A

```
[ERROR] == JReleaser ==
[ERROR] signing.pgp.secretKey must not be blank. Configure a value using the CLI yaml DSL, or define a System property jreleaser.gpg.secret.key, or define a JRELEASER_GPG_SECRET_KEY environment variable, or define a key/value pair in /Users/aalmiray/.jreleaser/config.toml with a key named JRELEASER_GPG_SECRET_KEY
[ERROR] Look for more details at /Users/aalmiray/dev/github/jreleaser/helloworld-java-bin/out/jreleaser/trace.log
JReleaser has not been properly configured.
signing.pgp.secretKey must not be blank. Configure a value using the CLI yaml DSL, or define a System property jreleaser.gpg.secret.key, or define a JRELEASER_GPG_SECRET_KEY environment variable, or define a key/value pair in /Users/aalmiray/.jreleaser/config.toml with a key named JRELEASER_GPG_SECRET_KEY
```

cc: @maxandersen

maxandersen:
## Problem

When `signing.pgp.active: ALWAYS` is configured but the PGP secret key is not available (e.g. `JRELEASER_GPG_SECRET_KEY` is not defined), running `jreleaser release --dry-run --yolo` fails with a validation error instead of skipping signing gracefully.

## Root Cause

In `SigningValidator.validateSigning()`, validation errors from `validatePgp()`, `validateCosign()`, and `validateMinisign()` were passed directly to the parent `errors` object, bypassing the `--yolo` mechanism that other validators (e.g. `AnnouncersValidator`) use.

## Fix

Applied the same `mergeErrors` pattern used by `AnnouncersValidator`:

1. Collect validation errors in a local `Errors incoming` object
2. After each sub-validator, call `mergeErrors(context, errors, incoming, signingTool)`
3. When `--yolo` is active, `mergeErrors` converts errors to warnings and calls `disable()` on the signing tool

## Tests

Added `SigningValidatorTest` with two tests:
- **`signingPgpValidationFailsWithoutYolo`** — verifies existing behavior: missing secret key produces configuration errors
- **`signingPgpValidationSkippedWithYolo`** — verifies fix: with `--yolo`, errors become warnings and PGP signing is disabled

Fixes #2128

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/signing/SigningValidator.java
@@ -46,6 +46,7 @@ import static org.jreleaser.model.api.signing.Signing.MINISIGN_PASSWORD;
 import static org.jreleaser.model.api.signing.Signing.MINISIGN_PUBLIC_KEY;
 import static org.jreleaser.model.api.signing.Signing.MINISIGN_SECRET_KEY;
 import static org.jreleaser.model.internal.validation.common.Validator.checkProperty;
+import static org.jreleaser.model.internal.validation.common.Validator.mergeErrors;
 import static org.jreleaser.model.internal.validation.common.Validator.resolveActivatable;
 import static org.jreleaser.util.CollectionUtils.listOf;
 import static org.jreleaser.util.Env.envKey;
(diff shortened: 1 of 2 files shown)


aalmiray (approved):

aalmiray:
Thank you 😄

PR #2140, opened 2026-06-09
fix: resolve current branch instead of assuming main

### LINKED ISSUE #2118: [release] JReleaser assumes main branch
When trying to make releases on github, jreleaser assumes the primary branch is called `main` and breaks on older repositories still using `master` as the primary branch.  See this log for an example:  https://github.com/MorphiaOrg/morphia/actions/runs/25965171984/job/76328059198#step:8:41886

### Environment Information

- **Operating System**: ubuntu
- **JReleaser Version:** 1.20.0
- **JDK Version:** 11

Please paste stacktraces from `[out|build|target]/jreleaser/trace.log`.

mvanhorn:
## Summary

The release branch defaulted to the literal string `main` whenever no branch was configured in the DSL or via the `JRELEASER_BRANCH` env var. Repositories whose default branch is `master` (or any non-`main` name) would fail the release because the assumed `main` branch does not exist.

This change resolves the default release branch from the current Git context (the checked-out HEAD ref already captured on the model's `Commit`) instead of assuming `main`, while keeping `main` as a last-resort fallback when no Git branch can be detected.

## Why this matters (#2118)

The reporter hit this on a `master`-based repository (MorphiaOrg/morphia). The documented workaround is to set the branch explicitly via the DSL or `JRELEASER_BRANCH`, but the hardcoded `main` default remains a footgun for the many older repositories that still use `master`. Resolving the actual checked-out branch removes that surprise without changing behavior for `main`-based repos.

## Changes

- `BaseReleaserValidator`: replace the unconditional `"main"` branch fallback with `resolveDefaultBranch(context)`, which returns the current branch (`context.getModel().getCommit().getRefName()`) when available and falls back to `"main"` only when no Git ref is detectable.
- Explicitly configured branch (DSL) and `JRELEASER_BRANCH`/property values continue to take precedence over detection, since they are still passed as the `value` argument to `checkProperty(...)`.
- The `branch.push` fallback chains off `service.getBranch()` and therefore inherits the corrected value automatically.

## Testing

Added `BaseReleaserValidatorTest` covering:
- repo on `master` with no branch configured resolves to `master`
- repo on `main` with no branch configured resolves to `main` (unchanged)
- an explicitly configured branch wins over detection
- no Git context falls back to `main`
- a blank ref name (e.g. detached HEAD) falls back to `main`

`./gradlew :jreleaser-model-impl:test --tests "*BaseReleaserValidatorTest"` passes (5/5).

Fixes #2118

### DIFF
--- a/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/release/BaseReleaserValidator.java
+++ b/core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/release/BaseReleaserValidator.java
@@ -137,7 +137,7 @@ public final class BaseReleaserValidator {
                 BRANCH,
                 baseKey + "branch",
                 service.getBranch(),
-                "main"));
+                resolveDefaultBranch(context)));
 
         service.setBranchPush(
             checkProperty(context,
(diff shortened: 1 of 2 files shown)


aalmiray:
Please read and abide to our contribution guidelines. 

No GenAI code is accepted. 

https://github.com/jreleaser/.github/blob/main/CONTRIBUTING.adoc

PR #2148, opened 2026-07-05
fix(brew): emit actionable hint when `JAVA_BINARY` distribution has `jar` artifact

### LINKED ISSUE #2124: [packagers] Brew packager skipped silently
### Task List

- [ ] Steps to reproduce provided
- [ ] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create a jreleaser configuration with a brew set-up in the `packagers` section and add a distribution of type `JAVA_BINARY` with a JAR-file instead of a ZIP-file. (This set-up is incorrect.)
2. Produce a JAR-file and refer to it in the distribution.
3. Run `jreleaser publish`.

### Expected Behaviour

It should fail with a message saying why the brew packager cannot be applied to the distribution and why. This is because the packager is in the `packagers` section and, hence, it would apply to all distributions.

### Actual Behaviour

The brew packager is skipped without a "disabled" message below the distribution in the log. Despite this, the generated configuration for the distribution shows the brew packager is disabled.

### Environment Information

- **Operating System**: macOS 26.5
- **JReleaser Version:** 1.24.0
- **JDK Version:** 21.0.8 (Oracle Corporation 21.0.8+12-LTS-250)

See also the discussion https://github.com/jreleaser/jreleaser/discussions/2123, where files have been uploaded.

HarshMehta112:
Fixes #2124

## Problem

  When a `JAVA_BINARY` distribution is configured with a `.jar` artifact and brew
  is set in the global `packagers` section, brew is silently disabled. The existing
  warning only says "no matching artifacts [.zip]" — it does not explain the
  mismatch or how to fix it.

  Reported in discussion #2123.

  ## Solution

  After the generic "no matching artifacts" warning, detect the specific case where
  a `JAVA_BINARY` distribution contains `.jar` artifacts and emit a second warning
  that names the problem and the two valid fixes:
  - switch distribution type to `SINGLE_JAR`
  - provide a `.zip` artifact instead

  ## Changes

  - `BrewPackagerValidator` — adds jar-on-JAVA_BINARY hint after generic disable warning
  - `Messages.properties` — adds `WARNING.validation.brew.java_binary.jar.hint` key
  - `BrewPackagerValidatorTest` — two new tests covering jar (hint emitted, packager
    disabled) and zip (no hint, normal validation path)

  ## Example warning output (after fix)

  distribution.myapp.brew was disabled because there are no matching artifacts [.zip]
  distribution.myapp has .jar artifacts but JAVA_BINARY distributions require .zip
  artifacts for brew packaging; switch the distribution type to SINGLE_JAR or
  provide a .zip artifact instead

### Checklist


aalmiray:
The PR description seems to be AI generated and thus I must ask, was AI also used for helping with the code?

HarshMehta112:
> The PR description seems to be AI generated and thus I must ask, was AI also used for helping with the code?

@aalmiray  Thanks for the feedback. Just to clarify, the code and unit tests were written entirely by me. I only used AI to improve the PR description based on my own comments for better clarity and structure.

Since the implementation itself is my own, I'm not sure why the PR description would prevent the code from being reviewed. If AI-assisted descriptions are a concern, I understand. Feel free to close the PR if needed. Thanks for your time.

aalmiray:
Thank you for clarifying. AI in code is a no go. AI on comments is still permissible as long as it does not result in a big pile of words.

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/packagers/BrewPackagerValidator.java
```diff
@@ -122,6 +122,13 @@ public static void validateBrew(JReleaserContext context, Distribution distribut
             context.getLogger().debug(RB.$("validation.disabled.no.artifacts"));
             errors.warning(RB.$("WARNING.validation.packager.no.artifacts", distribution.getName(),
                 packager.getType(), packager.getSupportedFileExtensions(distribution.getType())));
+            if (distribution.getType() == org.jreleaser.model.Distribution.DistributionType.JAVA_BINARY) {
```
aalmiray: I think the msg may be too specific for a single distribution type. I'd suggest using the `supportsDistribution(distributionType)` method to craft the msg

aalmiray (changes requested):

core/jreleaser-model-impl/src/main/java/org/jreleaser/model/internal/validation/packagers/BrewPackagerValidator.java
HarshMehta112: Makes sense. Implemented a generic fix using `supportsDistribution()` together with `getSupportedFileExtensions()` to display a warning message suggesting the compatible distribution type(s) when the selected distribution isn't supported.

HarshMehta112:
I messed up the local git history while applying changes, so I’ll close this PR and raise a new one with the correct updates.

PR #2158, opened 2026-08-01
feat(core): Add an option to override DSL values via envvars/sysprops

### LINKED ISSUE #2133: [core] Add an option to override DSL values via envvars/sysprops
Some properties that do not define an explicit value in the DSL may have their values resolved via envvars/sysprops. But if a value is explicitly defined in the DSL then it won't be overridden. 

Sometimes it would be desirable to externally override DSL values, avoiding duplicate config files. 

Provide an option to activate whether envvars/sysprops can override DSL values or nor, for example

```yaml
environment:
  override: true
```

**Describe alternatives you've considered**
Use different config files. 

**Additional context**
Suggested by @maxandersen

exoego:
Fixes #2133

### Context
Add an environment.override flag. When enabled, environment variables and system properties take precedence over values explicitly set in the DSL. System properties keep winning over environment variables.

The flag can also be turned on without touching the config file by setting JRELEASER_ENVIRONMENT_OVERRIDE or jreleaser.environment.override.

### Checklist
         https://github.com/jreleaser/jreleaser.github.io/pull/110

### DIFF
--- a/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/environment/Environment.java
+++ b/api/jreleaser-model-api/src/main/java/org/jreleaser/model/api/environment/Environment.java
@@ -27,9 +27,13 @@ import java.util.Properties;
  * @since 0.1.0
  */
 public interface Environment extends Domain {
+    String ENVIRONMENT_OVERRIDE = "ENVIRONMENT_OVERRIDE";
+
     Properties getVars();
 
     String getVariables();
+
+    boolean isOverride();
 
     Map<String, Object> getProperties();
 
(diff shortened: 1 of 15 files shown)


aalmiray:
This is quite a big change. Was AI used to generate code?

exoego:
Partly yes, I used LLM to come up with design and generate test codes, then changed production codes with Intellij IDEA to maintain my coding muscles, and iterated (changed tests and codes).

I also used LLM for the document update, so written English is oK enough. It is my idea to have an example for release-action with the new feature.

This level of code size is not so big for me, when I need the fix.

aalmiray:
OK. Just have to be extra sure that no direct Gen AI code is added to the codebase. We have a strict policy due to the gray area of licensing with these tools.

Issue #1074, opened 2022-11-30
[gradle] "clean" task raise a IOException because jreleaser directory cannot be deleted


joanhermay:
### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem (link to git repository is ideal)
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Have a clean gradle project (A Java project in the example and using Kotlin DSL)
3. Add the JReleaser plugin
4. Configure "jreleaser" section. The "project" section to be specific, in my case (and in the example): author, license, links and inceptionYear.
5. Run `.\gradlew build`
6. Run `.\gradlew clean`

### Expected Behaviour

`clean` task must delete `\build` directory as usual.

### Actual Behaviour

All folders in the `\build` directory are removed (`\build` folder is expected to be deleted at the end), except `\jreleaser` directory and it's content; a `IOException` is raised.

Following, the exception:

```
> Task :clean FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':clean'.
> java.io.IOException: Unable to delete directory 'C:\Users\joanhermay\Desktop\ioexception-in-gradlew-clean-task-due-to-jrelease-folder-cannot-be-deleted\lib\build'
    Failed to delete some children. This might happen because a process has files open or has its working directory set in the target directory.
    - C:\Users\joanhermay\Desktop\ioexception-in-gradlew-clean-task-due-to-jrelease-folder-cannot-be-deleted\lib\build\jreleaser\trace.log
    - C:\Users\joanhermay\Desktop\ioexception-in-gradlew-clean-task-due-to-jrelease-folder-cannot-be-deleted\lib\build\jreleaser
```

**NOTE:** The `\jreleaser` directory can be deleted if gradle daemond is killed.

### Environment Information

- **Operating System**: Windows 10 Pro 22H2 19045.2251
- **JReleaser Version:** 1.3.1
- **JDK Version:** 17.0.5

```
openjdk version "17.0.5" 2022-10-18 LTS
OpenJDK Runtime Environment Microsoft-6841604 (build 17.0.5+8-LTS)
OpenJDK 64-Bit Server VM Microsoft-6841604 (build 17.0.5+8-LTS, mixed mode, sharing)
```

- **Gradle Version:** 7.6
- **Use custom Gradle global properties?:** No
- **Use custom Gradle project properties?:** No

### Example that reproduces the problem

[...]

Issue #2150, opened 2026-07-07
[gradle] Use of 'Project.getProperties()' is not allowed with Isolated Projects


philliplbryant:
### Task List

- [X] Steps to reproduce provided
- [X] Stacktrace (if present) provided
- [X] Example that reproduces the problem (link to git repository is ideal)
- [X] Full description of the issue provided (see below)

### Steps to Reproduce

1. Execute any Gradle task from the `base` plugin (e.g. `gradlew clean` or `gradlew wrapper`)
2. Verify the tasks completes successfully
3. Execute `gradlew jreleaserConfig`
4. Verify exceptions:
```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':jreleaserConfig' (registered by plugin 'org.jreleaser').
> Unable to delete directory 'C:\Dev\projects\scm\isolated-project-reproducer\build\jreleaser'
    Failed to delete some children. This might happen because a process has files open or has its working directory set in the target directory.
    - C:\Dev\projects\scm\isolated-project-reproducer\build\jreleaser\trace.log

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights from a Build Scan (powered by Develocity).
> Get more help at https://help.gradle.org.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
Configuration cache problems found in this build.

2 problems were found storing the configuration cache, 1 of which seems unique.
- Unknown location: use of 'Project.getProperties()' is not allowed with Isolated Projects

See the complete report at file:///C:/Dev/projects/scm/isolated-project-reproducer/build/reports/configuration-cache/6zo2a687pld0xzi0nq5shnj3j/9nl6l6gyajeslvwc8achbot4r/configuration-cache-report.html
> Use of 'Project.getProperties()' is not allowed with Isolated Projects
```

### Expected Behaviour

The plugin should display the full configuration model for the project.

### Actual Behaviour

The task fails with two errors (I am new to JReleaser and am making Gradle specific changes, so I'm primarilry concerned with the second - please verify the JReleaser configuration to confirm an issue actually exists for the first):

1. Unable to delete directory 'C:\Dev\projects\scm\isolated-project-reproducer\build\jreleaser'
2. Problems were found storing the configuration cache ... Unknown location: use of 'Project.getProperties()' is not allowed with Isolated Projects

## Note:

[...]

Issue #2145, opened 2026-07-03
[gradle] JReleaser does not always create a trace log file in gradle


JordonPhillips:
`trace.log` is written only as a side effect of the shared `JReleaserLoggerService` build service's constructor, which is decoupled from task execution. As a result, a jreleaser task can run to completion (or fail) and write its other outputs while `trace.log`is never created. This is most painful on failure, because jreleaser directs the user to `build/jreleaser/trace.log` for details, even if it hasn't been created.

I'm not certain what exactly is causing the failure. I've looked into it a bit, but gradle code makes my head spin. I can say that I've observed that it seems related to gradle caching. A build in my repo with `--no-build-cache`, `--no-configuration-cache`, or `clean` seems to result in `trace.log` getting produced in the environment where I discovered the bug.

It's also worth noting that this behavior appeared after 1.23.0. In particular I noticed it on 1.25.0, though I can see there was a lot of refactoring of gradle tasks done for 1.24.0, so I suspect the cause is in there somewhere.

### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem (link to git repository is ideal)
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. create a simple gradle project with jrleaser 1.25.0
2. run `./gradlew --no-daemon jreleaserConfig`
3. run `ls build/jreleaser`

I've got a repro repo here: https://github.com/JordonPhillips/jreleaser-repro

That reproduces under slightly different circumstances than how I encountered the issue. I encountered it in CI where I unconditionally save the trace in case something goes wrong. I tried pushing that up so I could hand it off directly, but it's VERY sensitive so even so much as committing the various cache files killed the repro. I could give you a zip with the actual repro, but I understand that it may be difficult to trust a zip from a stranger. I can provide it if you like.

Also note that the repro that gets created is single use. As I've said it's very fragile and in this case the trace will start getting generated after the first run.

### Expected Behaviour

jreleaser should be unconditionally creating its trace log

### Actual Behaviour

jreleaser does not create its trace log

### Environment Information

- **Operating System**: macOS 26 (and ubuntu)
- **JReleaser Version:** 1.25.0
- **JDK Version:** 17

Please paste stacktraces from `[out|build|target]/jreleaser/trace.log`.

That's the funny part

aalmiray:
Before you make any progress, be aware that per our contribution guidelines we don't accept any AI assisted contributions.

[...]

Issue #771, opened 2022-02-22
JReleaser should support the Gradle Configuration Cache feature


bitspittle:
More here: https://docs.gradle.org/current/userguide/configuration_cache.html

You can test a project by running `gradle --configuration-cache help`.

When I was trying this out on my own project, I ran into this message: 

```
* What went wrong:
Configuration cache problems found in this build.

1 problem was found storing the configuration cache.
- Plugin 'org.jreleaser': registration of listener on 'Gradle.addBuildListener' is unsupported
  See https://docs.gradle.org/7.4/userguide/configuration_cache.html#config_cache:requirements:build_listeners

See the complete report at file:///home/d9n/Code/1p/kobweb/examples/helloworld/build/reports/configuration-cache/aie4zrn2laepsd7wcjs1fp0sf/u8pz4ts9wf2m4okvfuugzbw9/configuration-cache-report.html
> Listener registration 'Gradle.addBuildListener' by build 'kobweb' is unsupported.
```

I'm not 100% sure what JReleaser code looks like, but I recently changed some code in Kobweb that may help:

https://github.com/varabyte/kobweb/blob/main/gradle-plugins/application/src/main/kotlin/com/varabyte/kobweb/gradle/application/buildservices/KobwebTaskListener.kt

Here's the commit that introduced it: https://github.com/varabyte/kobweb/commit/161b227f978268b776322501d2aec2604f49d37a

For me, I was using the `addTaskExecutionListener` method to check if any task succeeded for failed. I was able to work around it by creating a service and using constructor injection to get access to a `buildEventsListenerRegistry` object: 

```
class KobwebApplicationPlugin @Inject constructor(
    private val buildEventsListenerRegistry: BuildEventsListenerRegistry
) : Plugin<Project>
   ...
   val service = project.gradle.sharedServices.registerIfAbsent("task-state-listener", KobwebTaskListener::class.java) {
        parameters { ... }
    }
   buildEventsListenerRegistry.onTaskCompletion(service)
```

Hope that helps. Good luck!

aalmiray:
Configuration cache is an incubating feature yet it fails the build when enabled and if it encounters issues ... will have to keep an eye on it.

bitspittle:
[...]

PR #2163, opened 2026-08-27
fix(gradle): Do not resolve the logger service during the configuration phase

philliplbryant:
Addresses one of the two failures reported in #2150.

### Context

Issue #2150 reports two independent failures from a single `gradlew jreleaserConfig` run.
This change fixes the first of them:

    Unable to delete directory '...\build\jreleaser'
    Failed to delete some children. This might happen because a process has files open
    or has its working directory set in the target directory.
    - ...\build\jreleaser\trace.log

The constructor of `AbstractJReleaserDefaultTask` resolved `JReleaserLoggerService` while
the task was being configured, only to obtain a logger for the debug output of
`JReleaserGradleProjectCapture`. Constructing that service opens
`build/jreleaser/trace.log`, and because `build/jreleaser` is the `@OutputDirectory` of
every JReleaser task, Gradle's stale output cleanup then tries to delete that directory
before running the task. On Windows the open file handle makes the deletion fail, so the
task cannot run.

The failure is not limited to the first build. The task never completes, so Gradle never
records the output directory and repeats the cleanup on every later build.

- `JReleaserGradleProjectCapture.groovy`: take the Gradle `Logger` instead of the
  `JReleaserLogger`, so capturing the project properties no longer needs the build service
- `AbstractJReleaserDefaultTask.groovy`: pass the task logger to
  `JReleaserGradleProjectCapture.of`

The build service is now created on first use inside the task action, after Gradle has
cleaned the output directory.

Closed issue #1074 covered the same open-handle mechanism, but only for the `clean` task,
which was addressed with a `doFirst` hook that closes the logger. That hook does not cover
Gradle's stale output cleanup of the `@OutputDirectory`.

### Verification

Sample single-project build on Windows, JDK 21.

On Gradle 9.6.1, the version reported in #2150: before this change `gradlew jreleaserConfig`
fails with the message above, and with `-Dorg.gradle.unsafe.isolated-projects=true` both of
#2150's failures appear together. After this change the delete failure is gone and only the
`Project.getProperties()` problem remains.

On Gradle 8.14.3: before this change `gradlew jreleaserChecksum` and
`gradlew jreleaserChecksum jreleaserConfig` both fail with the same delete error. After it,
both succeed.

### Follow-ups

Three further Gradle plugin fixes follow this one, each as its own pull request:

- Issue #771, configuration cache support. `JReleaserConfigTask` invokes `Task.project` at
  execution time, which the configuration cache does not allow.
- Issue #2145, `trace.log` not always written. The shared logger is closed several times
  during a build, and a closed `PrintWriter` silently swallows everything traced afterwards.
- Issue #2150, the remaining half of this report. `Project.getProperties()` is not allowed
  with Isolated Projects.

### Checklist
- [ ] Update [documentation](https://github.com/jreleaser/jreleaser.github.io) when applicable. — internal fix, no user-facing change

### DIFF
--- a/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/JReleaserGradleProjectCapture.groovy
+++ b/plugins/jreleaser-gradle-plugin/src/main/groovy/org/jreleaser/gradle/plugin/internal/JReleaserGradleProjectCapture.groovy
@@ -18,10 +18,10 @@
 package org.jreleaser.gradle.plugin.internal
 
 import org.gradle.api.Project
+import org.gradle.api.logging.Logger
 import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.plugins.JavaApplication
 import org.gradle.api.provider.Provider
-import org.jreleaser.logging.JReleaserLogger
 
 /**
  * Captures relevant information from a Gradle Project
@@ -70,11 +70,14 @@ class JReleaserGradleProjectCapture implements Serializable {
         'testResultsDirName'
     ]
 
-    static JReleaserGradleProjectCapture of(Project gradleProject, JReleaserLogger logger) {
+    // Takes the Gradle logger rather than the JReleaserLogger on purpose: the capture happens at
+    // configuration time, and resolving the JReleaserLoggerService here would open build/jreleaser/trace.log
+    // before Gradle gets a chance to clean that directory as a stale task output.
+    static JReleaserGradleProjectCapture of(Project gradleProject, Logger logger) {
         return new JReleaserGradleProjectCapture(gradleProject, logger)
     }
 
-    private JReleaserGradleProjectCapture(Project gradleProject, JReleaserLogger logger) {
+    private JReleaserGradleProjectCapture(Project gradleProject, Logger logger) {
         this.name = gradleProject.name
         this.group = gradleProject.group?.toString()
         this.version = gradleProject.version?.toString()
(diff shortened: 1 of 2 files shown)


aalmiray (commented):
Changes look good. I have to ask, did you use an AI assistant to generate code?

It's OK to use an AI assistant to understand the codebase, and write PR texts (as it seems it has been the case here). But we can't accept any lines of code that are provided by an AI assistant on licensing grounds.

philliplbryant:
> Changes look good. I have to ask, did you use an AI assistant to generate code?
> 
> It's OK to use an AI assistant to understand the codebase, and write PR texts (as it seems it has been the case here). But we can't accept any lines of code that are provided by an AI assistant on licensing grounds.

Yes, I used an AI assistant to generate code (and the PR text). I directed the work, tests, etc., and reviewed every change. But I did not actually write the code. I would have said so up front, but I didn't understand/realize the need to do so. 

I'm closing the pull request for these reasons. The reproducer that I previously submitted and everything in issue #2150 are my own work, unaided by AI. Unless you'd like for me to do otherwise, I'll leave them. And if you or someone else implements a fix, I'd be more than happy to help test it a multi-module build with configuration cache and isolated projects enabled.

My apologies for the misunderstanding.

philliplbryant:
Closed per earlier comment.

aalmiray:
@philliplbryant Thank you for your understanding. We have simple AI guidelines in our CONTRIBUTING doc. FWIW I followed the ideas presented in the text of this PR and committed https://github.com/jreleaser/jreleaser/commit/2fbce1924d7a3fee6fcf54b12cc4c206f285d154

philliplbryant:
> @philliplbryant Thank you for your understanding. We have simple AI guidelines in our CONTRIBUTING doc. FWIW I followed the ideas presented in the text of this PR and committed [2fbce19](https://github.com/jreleaser/jreleaser/commit/2fbce1924d7a3fee6fcf54b12cc4c206f285d154)

Much obliged.

Issue #1918, opened 2025-07-02
[deploy] XML resources crash when attempting Maven Central deployment


io7m:
### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [ ] Example that reproduces the problem (link to git repository is ideal)
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

I have a directory containing a set of artifacts to be deployed (the directory is laid out in the standard Maven repository layout). The directory includes artifacts that meet the CR requirements, along with PGP signatures for every file. I'm fairly sure, given the actual error below, that the artifacts themselves don't matter.

I have the following `jreleaser.toml`:

```
[deploy.maven.mavenCentral.sonatype]
active = "ALWAYS"
sign = false
url = "https://central.sonatype.com/api/v1/publisher"
stagingRepositories = ["build/maven"]
```

I run jreleaser:

```
$ env \
JRELEASER_MAVENCENTRAL_USERNAME="${MAVEN_CENTRAL_USERNAME}" \
JRELEASER_MAVENCENTRAL_PASSWORD="${MAVEN_CENTRAL_PASSWORD}" \
JRELEASER_PROJECT_JAVA_GROUP_ID="${JRELEASER_PROJECT_JAVA_GROUP_ID}" \
JRELEASER_PROJECT_VERSION="${JRELEASER_PROJECT_VERSION}" \
./jreleaser deploy --config-file=jreleaser.toml
```

### Expected Behaviour

I'd expect to see `jreleaser` create a bundle and deploy it to Maven Central.

### Actual Behaviour

```
[INFO]  JReleaser 1.19.0
[INFO]  Configuring with jreleaser.toml
[INFO]    - basedir set to /home/runner/work/android-maven-central-testing/android-maven-central-testing
[INFO]    - outputdir set to /home/runner/work/android-maven-central-testing/android-maven-central-testing/out/jreleaser
[INFO]  Reading configuration
[INFO]  git-root-search set to false
[INFO]  Loading variables from /home/runner/.jreleaser/config.properties
[WARN]  Variables source /home/runner/.jreleaser/config.properties does not exist
[INFO]  Validating configuration
[INFO]  Strict mode set to false
[INFO]  Project version set to 1.0.2
[INFO]  Release is not snapshot
[INFO]  Timestamp is 2025-07-02T11:54:01.450798Z
[INFO]  HEAD is at a384790
[INFO]  Platform is linux-x86_64
[...]

PR #2167, opened 2026-09-03
fix: Parse POMs without XPath so native-image works

### LINKED ISSUE #2166: GraalVM native image fails to parse POM and crashes
### Task List

- [x] Stacktrace (if present) provided

### Steps to Reproduce

1. Download https://github.com/jreleaser/jreleaser/releases/download/v1.26.0/jreleaser-native-1.26.0-osx-aarch64.zip, unpack and extract the bin
2. Create a staging repository holding one pom and one empty jar.

```sh
mkdir -p staging/com/example/demo/1.0.0

cat > staging/com/example/demo/1.0.0/demo-1.0.0.pom <<'POM'
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>demo</artifactId>
  <version>1.0.0</version>
  <packaging>jar</packaging>
</project>
POM

printf 'PK\005\006\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' \
  > staging/com/example/demo/1.0.0/demo-1.0.0.jar
```
3. Create `jreleaser.yml`.

```sh
cat > jreleaser.yml <<'YAML'
project:
  name: demo
  description: repro
  links:
    homepage: https://example.com
  authors:
    - nobody
  license: Apache-2.0
  inceptionYear: "2026"
  languages:
    java:
      groupId: com.example

signing:
  active: NEVER

deploy:
  maven:
    mavenCentral:
      demo:
        active: ALWAYS
        url: https://central.sonatype.com/api/v1/publisher
        applyMavenCentralRules: false
        sign: false
        checksums: false
        verifyPom: false
        stagingRepositories:
          - staging
YAML
```

4. Run it.

```sh
JRELEASER_MAVENCENTRAL_USERNAME=dummy \
JRELEASER_MAVENCENTRAL_PASSWORD=dummy \
JRELEASER_PROJECT_VERSION=1.0.0 \
./path/to/jreleaser deploy --dry-run
```

### Expected Behaviour

The pom is parsed successfully and a deployment bundle is assembled, as same as the JVM distribution does.

### Actual Behaviour

`java.lang.IllegalStateException: javax.xml.xpath.XPathExpressionException: java.lang.RuntimeException: exception creating new instance for pool` is thrown.
This error was already reported in #1918 but not fixed seemingly.

[trace.log](https://github.com/user-attachments/files/31763289/trace.log)

### Environment Information

- **Operating System**:  macOS 26.6.2 and Ubuntu (GitHub Action)
- **JReleaser Version:** 1.26.0
- **JDK Version:** None (since GraalVM native)

exoego:
Fixes #2166

### Context

`PathExpression.evaluate` reaches `StringBufferPool, which builds its `FastStringBuffer` instances reflectively. That constructor is not registered for reflection, so every POM parse fails in the native distribution with "exception creating new instance for pool".

Instead of maintaining `reflect-config.json` and bloating the binary file, this PR refactors the logic with plain DOM traversal, so XPath is not needed.

For unit testing, the private members are turned into package-private and static.
I avoided extracting `PomReader` or such to minimize code diff.

#### Disclosure

LLM was involved only in investigation. 
I written code manually.

#### Bonus

The XMLErrorResources bundle registered for #1918 only existed to produce that XPath error message, so it goes away too. 
Dropping XPath shrinks the native binary by 876 KB out of 98.7 MB.

### Checklist

### DIFF
--- a/sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
+++ b/sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
@@ -39,13 +39,13 @@ import org.jreleaser.util.CollectionUtils;
 import org.jreleaser.util.Errors;
 import org.jreleaser.util.FileUtils;
 import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.xpath.XPathConstants;
-import javax.xml.xpath.XPathExpressionException;
-import javax.xml.xpath.XPathFactory;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
@@ -163,7 +163,7 @@ public abstract class AbstractMavenDeployer<A extends org.jreleaser.model.api.de
         }
 
         try {
-            DeployableCollector collector = new DeployableCollector(root, context.getModel().getProject().isSnapshot());
+            DeployableCollector collector = new DeployableCollector(context, root, context.getModel().getProject().isSnapshot());
 
             java.nio.file.Files.walkFileTree(root, collector);
             if (collector.failed) {
@@ -672,13 +672,15 @@ public abstract class AbstractMavenDeployer<A extends org.jreleaser.model.api.de
         // noop
     }
 
-    private class DeployableCollector extends SimpleFileVisitor<Path> {
+    static class DeployableCollector extends SimpleFileVisitor<Path> {
+        private final JReleaserContext context;
         private final Path root;
         private final Set<Deployable> deployables = new TreeSet<>();
         private final boolean projectIsSnapshot;
         private boolean failed;
 
-        public DeployableCollector(Path root, boolean projectIsSnapshot) {
+        public DeployableCollector(JReleaserContext context, Path root, boolean projectIsSnapshot) {
+            this.context = context;
             this.root = root;
             this.projectIsSnapshot = projectIsSnapshot;
         }
@@ -711,7 +713,9 @@ public abstract class AbstractMavenDeployer<A extends org.jreleaser.model.api.de
             ));
         }
 
-        private PomResult parsePom(Path artifactPath) {
+        // Walks the document with plain DOM calls. XPath reaches JDK internals that are
+        // instantiated reflectively, which does not survive a GraalVM native-image build.
+        static PomResult parsePom(Path artifactPath) {
             // only inspect if artifactPath ends with .pom
             if (artifactPath.getFileName().toString().endsWith(EXT_JAR)) return new PomResult(PACKAGING_JAR);
             if (!artifactPath.getFileName().toString().endsWith(EXT_POM)) return new PomResult("");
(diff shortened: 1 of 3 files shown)


aalmiray:
Thank you for your contribution. Was any part of this code created using AI tools?

exoego:
No

LLM was involved only in investigation.
I written code manually.

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
```diff
@@ -722,25 +726,37 @@ private PomResult parsePom(Path artifactPath) {
...
-                    .evaluate(document, XPathConstants.STRING);
-
-                query = "/project/distributionManagement/relocation";
-                Object relocation = XPathFactory.newInstance()
-                    .newXPath()
-                    .compile(query)
-                    .evaluate(document, XPathConstants.NODE);
+
+                Element project = document.getDocumentElement();
+                if (!"project".equals(project.getTagName())) return new PomResult(PACKAGING_JAR);
```
aalmiray, line 731: Why is this required now?

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
```diff
@@ -39,13 +39,13 @@
 import org.jreleaser.util.Errors;
 import org.jreleaser.util.FileUtils;
 import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
```
aalmiray, line 44: Wouldn't these dom classes be required in the native-image reflection config now?

sdks/jreleaser-java-sdk-commons/src/test/java/org/jreleaser/sdk/commons/AbstractMavenDeployerTest.java
```diff
@@ -0,0 +1,134 @@
...
+            </project>
+            """);
+
+        var result = AbstractMavenDeployer.DeployableCollector.parsePom(path);
+
+        assertThat(result.relocated).isFalse();
+    }
+
+    @Test
+    void onlyDirectChildrenOfProjectAreConsidered() throws IOException {
```
aalmiray, line 104: Previous code did not handle this case, nor does the new one. Good to have a test case for it. Hope no pom like this is encountered in the wild, otherwise a fully resolved model would be required.

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
exoego: It is a guard to skip `<packaging>` out of `<project>`, the same way the XPath version did with `/project/packaging`.

XPath returned nothing when the root element was not `<project>`.

Plain DOM traversal would naively read children of whatever root it is given,
so a malformed `.pom` (unlikely, I guess) could yield a wrong packaging or a false relocation flag.

sdks/jreleaser-java-sdk-commons/src/main/java/org/jreleaser/sdk/commons/AbstractMavenDeployer.java
exoego: No worry.
The used DOM APIs are all statically resolved, so GraalVM native can include them.
While `XPathFactory.newInstance()` resolve XPath classes via reflection, which GraalVM needs reflection config.

sdks/jreleaser-java-sdk-commons/src/test/java/org/jreleaser/sdk/commons/AbstractMavenDeployerTest.java
exoego: Agreed, this pins existing behaviour rather than changing it.

On "in the wild", per the 4.0.0 and 4.1.0 XSDs 1<packaging>1 is not valid inside a `<profile>`, so that half of the fixture cannot occur.
`<distributionManagement>` is valid there, so a profile-scoped relocation is the only real case, and catching it would need the resolved model you mention.
Neither the fixture nor the `<project>` root check is about likelihood. Both pin what the XPath already did, so the change stays behaviour-preserving...

Happy to split the test to name the relocation case explicitly.

exoego: Splitted in https://github.com/jreleaser/jreleaser/pull/2167/changes/88c5cf444ce4b62d3ddc1c3e3628ca02a79dae0c
Is this ok?

aalmiray:
Thank you 😄

